PR #126, opened 2015-01-07
Removes some unused stuff.

himedlooff:
And helps @ascott1's PR to work better when starting from scratch.


ascott1:
Beautiful. Thanks for doing this @himedlooff

PR #132, opened 2015-01-16
Fixed typo in components.

wpears:
Though felxible sounds like a futuristic vehicle and maybe should be kept for the spirit of it


himedlooff:
Thanks! :smile:

PR #133, opened 2015-01-16
Some quicky starter info after running yo cf.

himedlooff:


himedlooff:
Wait for an update that addresses theming.

himedlooff:
Ok, added.

ascott1:
Thanks for this. It's going to be really useful for new folks.

PR #137, opened 2015-01-21
Added initial grunt build command

sephcoster:
Running simpleHTTPServer without the initial grunt build yields a blank page - this step (while indicated later, it's helpful to run it in line if you're running it the first time)


Scotchester:
Should probably run `grunt build-cf` before changing to the `dist` directory. (It'll work, but the only reason to be in `dist` is running the server, so it just seems to make the most sense to run it before going in there.)

sephcoster:
Updated to add the build-cf step prior to navigating to the dist directory.

Scotchester:
Thanks! :+1:

PR #147, opened 2015-01-29
Fixes #146 - adds info about warning dialog

### LINKED ISSUE #146: Add note in Readme about denying incoming connections for node?
CFPB laptops get the following warning when installing the project:
screen shot 2015-01-28 at 9 48 32 am

Do we want to include a message in the Readme about this? Or since this is only probably going to be an issue for our particular CFPB laptop configuration, is it not worth mentioning in this public readme?

Here's some suggested copy to add to the troubleshooting section in the readme:

```markdown
**Deny/Allow connections.** If you are an employee of the CFPB, you may encounter the following message:

> Do you want the application "node" to accept incoming network connections

Hit `Deny` if you encounter this dialog.
```

anselmbradford:
Fixes #146 - adds info about warning dialog box that asks whether node
should accept or deny incoming connections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,12 @@ The site should now be live at `http://localhost:4000/capital-framework/`.
 $ grunt vendor
 ```
 
+**Deny/Allow connections.** If you are an employee of the CFPB, you may encounter the following message:
+
+> Do you want the application "node" to accept incoming network connections
+
+Hit `Deny` if you encounter this dialog.
+
 ## Getting help
 
 If you encounter issues with the Capital Framework components shown in the documentation site, please post an issue in the Capital Framework [issue tracker](https://github.com/cfpb/capital-framework/issues/).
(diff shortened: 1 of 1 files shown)


anselmbradford:
Ready for review n merge!

README.md
```diff
@@ -39,6 +39,12 @@ The site should now be live at `http://localhost:4000/capital-framework/`.
 $ grunt vendor
 ```
 
+**Deny/Allow connections.** If you are an employee of the CFPB, you may encounter the following message:
```
Scotchester: I see no reason to make it CFPB-specific. I am fairly certain this dialog would pop up for anyone when they try to start a node server for the first time on an OS X installation.

anselmbradford: :+1: Sounds good, I've pushed an update. I'd not seen that dialog box before using the CFPB laptop, is it a Yosemite thing?

Scotchester: It happened on Mountain Lion, too. I'm guessing that on your personal machine or whatever, you saw it once, hit Allow or Deny, and it remembered your choice, because you had root privileges. Our users can't set the option permanently on these machines.

So, not a high likelihood that regular non-CFPB users will see it, but you never know.

Thanks!

Issue #31, opened 2014-03-05
Create releases with GitHub API/Gulp?


Scotchester:
As we want to distribute zip files for each release, we should think about using Grunt to create the releases via the GitHub API and eliminate the need to upload the file through the web interface.

This might leverage an existing task like [grunt-release](https://github.com/geddski/grunt-release), or it could be an extension of [grunt-cfpb-internal](https://github.com/cfpb/grunt-cfpb-internal). I probably favor the latter.

Scotchester:
I think this might still be worth doing. Renaming to reflect that CF is now built with Gulp, and moving out of VZ milestone.

jimmynotjim:
This has been automated, closing.

PR #149, opened 2015-03-26
Update hero instructions

jimmynotjim:
Updating hero instructions to match Generator instructions
 
## Additions
 
None
 
## Removals
 
None

## Changes
 
- added instructions from Generator README
 
## Testing
 
- None
 
## Review
 
- @Scotchester 
- @contolini 
- @ascott1 
 
## Preview
 
screen shot 2015-03-27 at 5 29 24 pm

## Notes
 
- Fixes [generator-cf #31](https://github.com/cfpb/generator-cf/issues/31)

### DIFF
--- a/_includes/hero.html
+++ b/_includes/hero.html
@@ -4,7 +4,9 @@
     <h2 class="hero_heading">A modular front end framework</h2>
 
 <!-- funky formatting to prevent indents -->
-<pre class="hero_code">$ npm install -g yo generator-cf
+<pre class="hero_code">
+$ npm install -g yo generator-cf
+$ mkdir my-new-project && cd $_
 $ yo cf
 </pre>
 


ascott1:
Looks good. I might nix the `open ./dist/index.html` bit, but I can be convinced otherwise.

jimmynotjim:
@ascott1 I included it in the generator README because I feel like without it you're just left wondering "now what" when you're new to the project. Taking a look at the index file allows you to sort of "kick the tires" and get a feel for how things work, and from there you can get working. 

I also wonder if we shouldn't explain that the index file is copied from `src/` because that could suck if you're markup work in `dist/`was overwritten upon building.

Overall, I've tried to apply viewpoint of a new user to all the changes I've submitted this week.

ascott1:
Makes sense. To me "quick start" implies that you're comfortable digging in a bit, but we should be as explicit as possible on the "getting started" page.

jimmynotjim:
Ah, I forgot about the "getting started" page. Maybe we just need to call that out a bit more? Right now it's a bit lost with the other two lines on dependencies and compiled source. 

If we were to do that I think I could remove 
```
$ grunt
$ open ./dist/index.html
```

leaving
```
$ npm install -g yo generator-cf
$ mkdir my-new-project && cd $_
$ yo cf
```

in the hero.

ascott1:
:+1: I think that's good and brief for the hero and we can add those explicit instructions to the getting started page. 

I agree that we should probably call that page out a bit more. It's a teeny tiny link.

jimmynotjim:
All updated. If this is good I'll squash it down for merging.

Scotchester:
:+1:

jimmynotjim:
Squashed and ready

PR #166, opened 2015-05-13
Download Button

ascott1:
This adds a big 'ol "Download" button to the hero. I feel like we need some simple, but explanatory copy here as well to clarify the Yeoman generator vs the download button. Thoughts @contolini @Scotchester ?

screen shot 2015-05-13 at 2 18 07 pm

### DIFF
--- a/_includes/hero.html
+++ b/_includes/hero.html
@@ -10,9 +10,7 @@ $ mkdir my-new-project && cd $_
 $ yo cf
 </pre>
 
-    <p class="hero_p sans">You will need <a href="http://nodejs.org/">Node.js</a>, <a href="http://yeoman.io/">Yeoman</a>, <a href="http://gruntjs.com/">Grunt</a>, and <a href="http://bower.io/">Bower</a> to run these commands.</p>
-    <p class="hero_p sans">Need more detailed instructons? <a class="jump-link" href="{{ site.baseurl }}/getting-started/">Read the Getting Started guide </a></p>
-    <p class="hero_p sans">Not into the command line? <a class="jump-link jump-link__download" href="assets/cf.zip">Download the compiled source </a></p>
-
+    <a href="assets/cf.zip" class="btn btn__super">Download <span class="btn_icon__right cf-icon cf-icon-download"></span></a>
+    <a href="{{ site.baseurl }}/getting-started/" class="btn btn__super">Get Started <span class="btn_icon__right cf-icon cf-icon-right"></span></a>
   </div> <!-- /.wrapper -->
 </div> <!-- /.hero -->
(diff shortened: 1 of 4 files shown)


contolini:
:+1: I agree we need to clarify that it's a "do this *or* that but not both" scenario. Maybe we could just put the "Not into the command line?" text back in front of the button?

[Cleverstack](http://cleverstack.io/) has an interesting way of handling it. They ask if `npm install` means anything to the user.

ascott1:
The tricky thing with "Not into the command line?" is disassociating it from the "Get Started" button.

contolini:
Any thoughts @cfpb/front-end-team-admin? Seems like we're almost there, it just needs a little UX massaging.

Scotchester:
How about moving the download button to the space to the right of the command prompt? And I would suggest restoring the "Need more detailed instructions?" copy to pair with the "Get Started" button.

Scotchester:
To be explicit about something that was in my head but forgot to write: show the word "or" between the command prompt and the download button.

ascott1:
> To be explicit about something that was in my head but forgot to write: show the word "or" between the command prompt and the download button.

This is the tricky design bit. It feels weird... let me play around with this over the next couple of days to see what I can come up with.

awolfe76:
I think part of the weird feeling is that the "getting started" points to how to get started with the generator.

Maybe instead of highlighting the generator on the home page, talk about "CF offers a variety of methods to get started depending on use cases".

3 methods:

1. Using the generator
  - Still link to getting started but maybe re-word to "getting started with the generator".
1. Downloading compiled CSS
  - In addition to the download button, add a link to the components section; or (more work required) modify the getting started page to include a "getting started with compiled CSS" with example of `<link href="your/directory/cf.css">` and `<script src="your/directory/file.js">` and link to that anchor point.
1. Downloading source
  - With some language about tools needed to build yourself and a link to, roughly, the current "Navigating the project folder" part of the getting started page.

ascott1:
I think that's great feedback @awolfe76 

I didn't forget about this. The week has just been getting away from me... I tried pinging a few of our designers for help, but haven't heard back yet.

ascott1:
Closing this based on a conversation @contolini and I had. Our new approach is going to be:

- kill the "You will need Node.js, Yeoman, Grunt, and Bower to run these commands." line
- put the download sentence right underneath the yeoman commands and make it bigger
- and underneath that, "Need more detailed instructons? Read the Getting Started guide"

contolini:
Eerily, I submitted a PR (https://github.com/cfpb/capital-framework/pull/177) with changes incredibly similar to what @awolfe76 suggested above *before* reading this thread! :boom:

PR #174, opened 2015-06-05
Restructure the entire thing

contolini:
This PR updates this repo to use our standardized best practices. Fixes #148 and #143.

## Changes
- `assets/` and `build/` dirs changes to `static/` and `dist/`. `src/` was not used because `capital-framework.less` is in that directory.
- Gruntfile overhaul.
- Location of archive moved from `assets/cf.zip` to `releases/capital-framework-x.x.x.zip` *and* `releases/capital-framework-latest.zip`
- Lots of general tidying.

## Review

- @Scotchester 
- @ascott1 
- @himedlooff

PR #177, opened 2015-06-05
Restructure dist dir and clarify how to get started

contolini:
## Changes

- Clarifies how to use CF on the getting-started page. ([Preview](https://github.com/contolini/capital-framework/blob/restructure/getting-started/index.md))
- Restructures dist dir to match https://github.com/cfpb/capital-framework/pull/174#issuecomment-109352958.
- Rebuilds archives.

## Review

- @ascott1 
- @Scotchester 

## Preview

```
dist/
├── capital-framework.css
├── capital-framework.ie.css
├── capital-framework.ie.min.css
├── capital-framework.js
├── capital-framework.min.css
├── capital-framework.min.js
└── demo-app
    ├── README.md
    ├── index.html
    └── static
        ├── css
        │   ├── main.css
        │   ├── main.ie.css
        │   ├── main.ie.min.css
        │   └── main.min.css
        ├── fonts
        │   ├── cf-icons.eot
        │   ├── cf-icons.svg
        │   ├── cf-icons.ttf
        │   └── cf-icons.woff
        ├── js
        │   ├── main.js
        │   ├── main.min.js
        │   └── main.min.js.map
        └── vendor
            └── box-sizing-polyfill
                └── boxsizing.htc
```

## Notes

I included a README in the demo-app dir that tells you how to start a local server and to use the generator if you want a better sample application.


dist/capital-framework.css
```diff
@@ -4915,6 +4915,191 @@ input.btn::-moz-focus-inner {
    Capital Framework
    Form Element Styling
    ========================================================================== */
+/* ==========================================================================
+   Capital Framework
+   Grid mixins
+   ========================================================================== */
```
Scotchester: Any idea how all this cf-grid stuff got moved inside the cf-forms CSS?

dist/demo-app/README.md
```diff
@@ -1,74 +1,27 @@
 # Capital Framework
 
-A Capital Framework demo project. Use grunt to process the source files in `src/` and output them to `dist/`:
+A Capital Framework sample project.
+To preview it, run `python -m SimpleHTTPServer` and open http://localhost:8000 in a web browser.
```
Scotchester: Keeping in mind the type of user likely to download the precompiled archive, maybe also mention that they should just be able to open the `index.html` file from their file system in a browser?

dist/capital-framework.css
contolini: That's an artifact of the new `@import` build process. The Less compiler is concatenating the stylesheets and because the `@import` rules are [below the cf-forms banner](https://github.com/cfpb/cf-forms/blob/gh-pages/src/cf-forms.less#L6-L9), that's where they're injected. Seems like I probably should have put them at the very top of every components' stylesheet, *above* the banner. :expressionless:

contolini: I'll patch the components but I'm going to need some coffee first...

dist/demo-app/README.md
contolini: That won't work because index.html uses absolute paths for the stylesheets... which we probably shouldn't be doing. I'll switch them to relative URLs.

Scotchester: Mmm, good point... Is `index.html` file the same file as `src/index.html` that comes out of the generator? Or is it customized for this downloadable archive?

contolini: It *was* directly from the generator. But now that we only include the built assets in demo-app, I'm okay with it being different from the generator output. The only times we'll need to update it are when we introduce new components or make breaking changes and neither scenario is very common.

Scotchester: :cool:

contolini:
@Scotchester @himedlooff That latest commit adds a second Less task to generate the CSS for the archive. There's a new [src file](https://github.com/contolini/capital-framework/blob/3936305095cd5f585b3d85353f48af30bb861540/src/capital-framework-generated.less) that links to cf-grid's `cf-grid-generated.less` to properly generate all the grid CSS classes.

Things will become a lot clearer after we split this repo into two branches, but I won't have time to get to that today.

Scotchester:
Looks good to me, now :+1:

ascott1:
Ditto. Merging.

PR #187, opened 2015-09-29
fix input-disabled var

cfarm:
Fixes the value for input-disabled var - should be gray-10 instead of gray-50.

dependency for cf-forms: https://github.com/cfpb/cf-forms/pull/37


wpears:
:exclamation:

PR #247, opened 2015-12-18
Version bump for latest in cf-layout

jimmynotjim:
Updated the dependency list for the latest release of cf-layout

## Additions

- Added missing CHANGELOG.md

## Changes

- Bumped version in bower.json
- Bumped version in package.json

## Testing

- ¯\_(ツ)_/¯ 

## Review

- @contolini 
- @KimberlyMunoz 
- @cfarm 
- @Scotchester 

## Screenshots

:us: 

## Checklist

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "capital-framework",
-  "version": "1.0.0",
+  "version": "2.0.0",
   "description": "Project-wide documentation for Capital Framework.",
   "homepage": "http://cfpb.github.io/capital-framework/",
   "author": {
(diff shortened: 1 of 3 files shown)


contolini:
:+1:

bower.json
```diff
@@ -23,7 +23,7 @@
     "cf-icons": "^1.0.0",
     "cf-pagination": "^1.0.0",
     "cf-typography": "^1.0.0",
-    "cf-layout": "^1.0.0"
+    "cf-layout": "^2.0.0"
```
Scotchester: Why not update any of the other dependency versions?

contolini: Only cf-layout moved to 2.x.x due to breaking changes. Everything else is still 1.x.x.

Scotchester: Yeah, but the others that were already over 1.0.0 before the 1.0.0 push have more specific version numbers.

jimmynotjim: The last discussion we had in the FEWD call was that individual modules would keep their own versions and only CF itself would update whenever one of the modules did. I think that's the smartest route because someone installing just cf-core, shouldn't need to update their deps when we release the next patch to cf-core.

PR #256, opened 2016-01-04
Update Less' paths option

contolini:
@Scotchester these are the changes we walked through last week.

## Changes
- Gulp styles task to use Less' paths option instead of npm import plugin.
- Normalize.css @import paths in cf-core and cf-grid to play nicer with Less paths option.
- Added fresh copies of icon fonts to fix "Failed to decode downloaded font" error.

## Review
- @Scotchester 

## Notes

- After this is merged, Travis will (hopefully) build and publish cf-core, cf-grid and capital-framework to npm.


CHANGELOG.md
```diff
@@ -3,14 +3,10 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
 ## Unreleased
 
-### Added
-- 
-
```
Scotchester: This should have a header for 3.0.9 and the placeholders for Unreleased should stay in place.

CFPBot: Travis'll automagically handle that.

contolini: Travis'll automagically handle that.

contolini: To clarify: As part of the pre-publishing process, Travis takes everything under the "Unreleased" heading and adds a version number and timestamp above it.

Scotchester: Mmmkay. Seems a little too magical to me, but let's see how it plays out.

Scotchester:
Working for me locally. :+1:

PR #258, opened 2016-01-04
Improve contributing docs in readme

contolini:
@Scotchester 

This is a hotfix test. This PR is opened against master instead of canary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,56 @@ npm install cf-buttons
 npm install cf-expandables
 ```
 
+Want some boilerplate?
+
+```sh
+npm install -g yo generator-cf
+mkdir my-new-project && cd $_
+yo cf
+```
+
 | buttons | core | expandables | forms | grid | icons | layout | pagination | typography |
 |:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -13,10 +13,57 @@ npm install cf-buttons
...
 
+## Contributing
+
+### Building components locally
+
+```sh
+git clone git@github.com:cfpb/capital-framework.git
+cd capital-framework
+npm install
+npm run build
```
Scotchester: Not `gulp build`?

contolini: `npm run build` runs `gulp build` [behind the scenes](https://github.com/cfpb/capital-framework/blob/canary/package.json#L27). Right now they're the same thing but it's good practice to tell people to use the npm script because in the future we might add another build step before or after `gulp build`. Or, who knows, we might move back to grunt. This way the build command stays the same.

Scotchester:
:+1: Here we go!

README.md
KimberlyMunoz: I love this. I opened an issue in the standards to see how folks feel about using it on other projects here: https://github.com/cfpb/front-end/issues/101

:can-of-worms:

PR #260, opened 2016-01-11
Remove block borders when __flush modifers used

Scotchester:
Removes corresponding borders on `block`s when `__flush` modifiers are also applied.

## Testing

1. Pull changes.
2. Run `npm link` from `src/cf-core` directory.
3. Run `npm install` in one of your projects to update cf-core.
4. View and confirm correct list item spacing.

## Review

- @cfpb/front-end-team-admin 

## Checklist

### DIFF
--- a/src/cf-layout/src/cf-layout.less
+++ b/src/cf-layout/src/cf-layout.less
@@ -312,7 +312,7 @@
 .content__flush-all-on-small {
     .respond-to-max(@bp-sm-max, {
         padding: 0;
-        border-width: 0;
+        border: none;
     });
 }
 
--- a/src/cf-layout/src/cf-layout.less
+++ b/src/cf-layout/src/cf-layout.less
@@ -371,7 +371,7 @@
             margin-left: -@grid_gutter-width;
         });
 
-        respond-to-max(@bp-xs-max, {
+        .respond-to-max(@bp-xs-max, {
             &.block__border,
             &.block__border-right,
             &.block__border-left {
--- a/src/cf-layout/src/cf-layout.less
+++ b/src/cf-layout/src/cf-layout.less
@@ -366,18 +366,18 @@
         margin-right: -(@grid_gutter-width / 2);
         margin-left: -(@grid_gutter-width / 2);
 
-        .respond-to-min(@bp-sm-min, {
-            margin-right: -@grid_gutter-width;
-            margin-left: -@grid_gutter-width;
-        });
-
         .respond-to-max(@bp-xs-max, {
             &.block__border,
             &.block__border-right,
             &.block__border-left {
-                border-right: none;
-                border-left: none;
+              border-right: none;
+              border-left: none;
             }
+        });
+
+        .respond-to-min(@bp-sm-min, {
+            margin-right: -@grid_gutter-width;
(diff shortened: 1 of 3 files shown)


src/cf-layout/src/cf-layout.less
```diff
@@ -15,12 +15,12 @@
 // Color variables
 
 // .block
+@block__bg:                     lighten(#3a8899, 55%);
 @block__border-top:             #3a8899;
 @block__border-right:           #3a8899;
 @block__border-bottom:          #3a8899;
 @block__border-left:            #3a8899;
 @block__border:                 #3a8899;
```
jimmynotjim: Should `@block__border` be moved up as well? Then each side can reference it so if it changes, it's only changing once.

src/cf-layout/src/cf-layout.less
```diff
@@ -360,6 +370,15 @@
             margin-right: -@grid_gutter-width;
             margin-left: -@grid_gutter-width;
         });
+
+        .respond-to-max(@bp-xs-max, {
```
jimmynotjim: Why do these need to be in a media query but the ones on L:394 don't?

src/cf-layout/src/cf-layout.less
Scotchester: Oops, yeah. Good call.

src/cf-layout/src/cf-layout.less
Scotchester: Because I'm stupid. Thanks for catching :)

Scotchester:
@jimmynotjim Issues fixed. Thanks!

src/cf-layout/src/cf-layout.less
```diff
@@ -346,16 +346,35 @@
...
+        &.block__border-bottom {
+            border-bottom: none;
+        }
     }
 
     &__flush-sides {
         margin-right: -(@grid_gutter-width / 2);
         margin-left: -(@grid_gutter-width / 2);
 
+        .respond-to-max(@bp-xs-max, {
```
jimmynotjim: Sorry, what I meant was do we need the media query at all? I'm failing to understand why we would want the borders back on a flush-sides after it reaches the `@bp-sm-min` breakpoint. Possibly related, should this be updated to `none`? https://github.com/cfpb/capital-framework/pull/260/files#diff-5140a128023e2d3dcfd06c7e5b80e2c5R315

Scotchester: Because the page has max-width, so it's not truly "flush". We still want borders on the sides. (If there's a sidebar, the way we're drawing the sidebar border places it overtop of the block's border.)

Check out a couple live examples on this page: http://www.consumerfinance.gov/adult-financial-education/

As to the think you linked: Maybe? Typically `border-width: 0` and `border: none` have the same effect.

jimmynotjim: Ah ok, thanks for clearing that up. Mind updating the one I linked to use `none` so it's consistent?

Scotchester: Done and ready for final review.

jimmynotjim:
Dropped the ball, apologies. :+1:

PR #271, opened 2016-02-11
Adding font-size vars for consistent sizing

jimmynotjim:
Adding font-size vars for consistent sizing

## Additions

- Added new variables matched to the corresponding heading level.

## Changes

- Updated `heading-#` mixins to use new variables in place of px values.

## Testing

- Source locally with `npm link` and build your project, headings should all be the same size.

## Review

- @KimberlyMunoz 
- @Scotchester 
- @cfarm 
- @contolini 

## Notes

- Now easier to use font-sizes without exposing the `@font-size` var.
- Naming based on http://csswizardry.com/2012/02/pragmatic-practical-font-sizing-in-css/.
- Adds `@size-giga`, `@size-mega`, `@size-milli`, `@size-micro` for future use.
- Fixes [cfgov-refresh #1387](https://github.com/cfpb/cfgov-refresh/issues/1387)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,13 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 ## Unreleased
 
 ### Added
-- 
+- [MINOR] **cf-core** Added font-size vars
 
 ### Changed
-- 
+-
 
 ### Removed
-- 
+-
 
 ## 3.1.2 - 2016-02-10
 
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -66,7 +66,7 @@ body {
     line-height: @base-line-height;
 }
 
-.heading-1( @fs: 34px ) {
+.heading-1( @fs: @size-alpha ) {
     @font-size: @fs;
     .webfont-regular();
 
@@ -75,7 +75,7 @@ body {
     line-height: 1.25;
 }
 
-.heading-2( @fs: 26px ) {
+.heading-2( @fs: @size-beta ) {
     @font-size: @fs;
     .webfont-regular();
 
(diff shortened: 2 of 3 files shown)


Scotchester:
I don't know that I love that naming scheme, even after reading the well reasoned article on it. I need to stew on it.

jimmynotjim:
My biggest support of it, is there are times a mockup or discussion will reference text as say an H4, but only mean the size, not the weight change (like the lead paragraph). Separating the sizing names from the rest of the heading styles will help break that confusion.

anselmbradford:
How come there’s a mix of metric prefixes and not? Could you use metric prefixes across the board? That might be more familiar https://en.wikipedia.org/wiki/Metric_prefix

jimmynotjim:
I'm on board with that approach as well. Whatever gets us discussing sizes and heading styles distinctly works for me.

KimberlyMunoz:
Would something like roman numerals work better so we still have a little obfuscation to show this doesn't have the responsive behavior of the h[1-6] classes, but developers don't have to memorize a new number scale to find the right font-size variable?

jimmynotjim:
I dig it, updated.

Scotchester:
You didn't update the vars in `cf-base.less`.

How about `size-lg` for the superheading font size variable? So it's not misinterpreted as Roman numeral 50, or as Arabic numeral one.

Scotchester:
P.S.: I like the Roman numeral approach.

jimmynotjim:
What in the world? I updated `cf-base`, no idea why it didn't push through because it's updated locally. Will figure that out, but in the mean time, `l` was intentional as 50, because it's crazy high. Is `x` better so we stick to the naming convention we've chosen?

Scotchester:
@jimmynotjim The problem is that the Roman numerals grow larger as the size decreases. `L` would seem to indicate 1px type ;)

jimmynotjim:
Ah, right, I hadn't thought of that. Updated to `xl` because it should be clear it's giant.

Scotchester:
:+1:

PR #285, opened 2016-02-23
keep usage.md file on component build

ascott1:
This copies the `usage.md` file into the component on build so that it can be consumed by the documentation site. This is the simplest solution, but it may be preferable to put this file in the `src` directory.

## Testing

Pull down this branch, run `npm run build` and take a gander at the output in the `tmp` directory.

## Review

Any combination of the following:

- @jimmynotjim 
- @contolini
- @cfarm 
- @KimberlyMunoz 
- @Scotchester

### DIFF
--- a/scripts/gulp/copy.js
+++ b/scripts/gulp/copy.js
@@ -23,7 +23,6 @@ gulp.task( 'copy:components:source', function() {
       var component = file.path.split('/').pop(),
           src = [
                   file.path + '/**',
-                  '!' + file.path + '/usage.md',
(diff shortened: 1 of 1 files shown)


ascott1:
The goal is to be able to bring the raw `usage.md` files into the build process for the documentation site.

To do this, I plan to output each of them into a Jekyll template with the markup. That will all happen on the `gh-pages` branch, but I first need access to the raw markdown.

Scotchester:
:+1:

PR #319, opened 2016-05-26
Automatically update compiled source links and re-compile component docs

contolini:
Re-compiled the docs and added a script to pull the latest version of CF and modify every link with a class of `cf-download` to point to that latest version.

@Scotchester


src/static/js/main.js
```diff
@@ -19,3 +19,11 @@ $( '.feature-list' ).append(
   'cf-components.</p>' +
   '</section>'
 );
```
Scotchester: You need to remove this obsolete `feature-list` counting stuff. It's throwing a console error and stopping the execution of the new script that you added below.

contolini: D'oh. Thanks.

Scotchester:
:+1:

PR #359, opened 2016-07-25
Removes vendor prefixes handled by autoprefixer

anselmbradford:
## Changes

- Removes vendor prefixes handled by autoprefixer.

## Testing

- Place `outline: 1px solid green;` after the styles that have vendor prefixes removed.
- Run `gulp styles`
- Check `dist/capital-framework.css` and search for `green` and note that styles have vendor prefixes added.

## Review

- @Scotchester 
- @jimmynotjim 
- @KimberlyMunoz
- @sebworks
- @contolini

## Notes

- `transform` does not add `-ms-transform`, indicating this is for IE older than 8, which should be covered by `filter` anyway, so I do not think this is needed.

### DIFF
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -556,14 +556,13 @@ select[multiple] {
     border: 1px solid @input-border;
     border-radius: 0;
     vertical-align: top;
-    -webkit-appearance: none;
-    -webkit-user-modify: read-write-plaintext-only;
+    appearance: none;
 }
 
 // Overrides extra left padding.
 // http://stackoverflow.com/questions/11127891/how-can-i-get-rid-of-horizontal-padding-or-indent-in-html5-search-inputs-in-webk
 ::-webkit-search-decoration {
-    -webkit-appearance: none;
+    appearance: none;
 }
 
 // The .focus class is only included for documentation demos and should not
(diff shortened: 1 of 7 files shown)


jimmynotjim:
Looks good 👍

src/cf-core/src/cf-base.less
```diff
@@ -556,14 +556,16 @@ select[multiple] {
     border: 1px solid @input-border;
     border-radius: 0;
     vertical-align: top;
-    -webkit-appearance: none;
+    appearance: none;
+    // TODO: update to use CSS 3 Basic UI specs as user-modify is deprecated.
     -webkit-user-modify: read-write-plaintext-only;
+    -moz-user-modify: read-write-plaintext-only;
```
Scotchester: 1. Why not just remove this now?
2. `-moz-user-modify` [doesn't support that value](https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-user-modify)

anselmbradford: Oops, you're right, I misread that footnote in the docs. Sure, I can remove them both now.

anselmbradford: Removed in https://github.com/cfpb/capital-framework/pull/359/commits/6909fc553a1e1c33b7fa05604f5f383a620976df

Scotchester:
:+1:

PR #326, opened 2016-06-21
Updating cf-pagination to use the atomic structure from cfgov-refresh

jimmynotjim:
This is the first update using the atomic code organization. It serves as an example of how each component should be converted.

## Additions

- Pagination molecule stylesheet

## Changes

- Updated documentation to show updated code examples
- Usage file for gh-pages

## Testing

- Install CF Sandbox https://github.com/jimmynotjim/capital-framework-sandbox#installation
- Click around to ensure the local server is rendering the sandbox correctly
- Stop the server (ctrl+c)
- In your local CF repo, checkout this branch and run `npm run cf-link`
- In your local CF Sandbox checkout the `cf-pagination-example` branch
- In your local CF Sandbox run `npm run cf-link && npm run build && npm start`
- Navigate to `http://localhost:3000/components/cf-pagination/`
- The _Original Documentation Markup_ should look mostly unstyled, the _Atomic Documentation Markup_ should look like the screenshot below, the _Live CFGov Markup_ should look completely broken (this is expected at the moment).
- To revert the cf-pagination changes in your local CF Sandbox run `npm unlink cf-pagination && npm install cf-pagination`

## Review

- @cfpb/front-end-team-admin 

## Screenshots

screen shot 2016-06-21 at 4 48 57 pm

## Notes

- It's difficult to test live code at the moment because these styles already exist in cfgov-refresh and testing in another project requires updating the project's markup. ~~A sandbox to test in is coming to make things easier.~~ It's happening!!!! https://jimmynotjim.github.io/capital-framework-sandbox/components/cf-pagination/
- `npm link` can be weird some times. I've found even when I have the new linked module in my `node_modules` dir I will still get the old styles when running `gulp build`. Making sure the local server isn't running and running the build task a few times seems, then restarting the server seems to fix it. ¯\_(ツ)_/¯ 
- The Sandbox currently includes base examples of cf-vars, cf-expandables, cf-pagination, and cf-buttons. All the other components need to have base examples included, and then each one can be updated with atomic examples as they are picked up.
- To find the atomic versions in cfgov-refresh, navigate to `cfgov/unprocessed/css` and check the `/atoms`, `/molecules`, and `/organisms` folders. Also be sure to check `cf-enhancements` for one-off updates that may or may not need to be included in the conversion.

## Todos

__ATOMIZE ALL THE THINGS!__

shrunken-kids

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
[...]

PR #367, opened 2016-07-29
Convert cf-core to Atomic Design conventions

Scotchester:
This PR updates the few parts of cf-core that need updating to follow Atomic Design conventions.

## Changes

  - `.webfont-<style>()` mixins renamed to `.u-webfont-<style>()`
  - `body` font is now `@webfont-regular` (Arial, by default).

## Removals

- Deprecated items:
  - `@mobile-max`
  - `@tablet-min`
  - `.subheader`
  - `.superheader`
  - `.figure__bordered`

## Testing

- See instructions in #326 

## Review

- @jimmynotjim 
- @KimberlyMunoz 

## Todos

- Add `npm run link` support after #326 is merged.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "postinstall": "node scripts/npm/postinstall",
     "release": "./scripts/travis/release.sh",
     "build": "gulp build",
-    "test": "gulp test && node test/accessibility/wcag.js"
+    "test": "gulp test && node test/accessibility/wcag.js",
+    "cf-link": "cd ./src/cf-core && npm link && cd ../cf-buttons/ && npm link"
   },
   "devDependencies": {
     "bluebird": "^3.0.5",
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -539,7 +539,7 @@ select[multiple] {
     @font-size: 16px;
 
     display: inline-block;
-    padding: unit( 6 / @base-font-size-px, em );
+    padding: unit( 6px / @base-font-size-px, em );
     border: 1px solid @input-border;
     border-radius: 0;
     margin: 0;
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -8,48 +8,51 @@
 //
 
 @webfont-regular: Arial;
-@webfont-italic: Arial;
-@webfont-medium: Arial;
-@webfont-demi: Arial;
+@webfont-italic:  Arial;
+@webfont-medium:  Arial;
+@webfont-demi:    Arial;
 
-.webfont-regular() {
+.u-webfont-regular() {
     font-family: @webfont-regular, Arial, sans-serif;
     font-style: normal;
     font-weight: normal;
 
     & em,
     & i {
-        .webfont-italic();
+        .u-webfont-italic();
     }
 
     & strong,
     & b {
-        .webfont-demi();
+        .u-webfont-demi();
     }
 }
 
-.webfont-italic() {
+.u-webfont-italic() {
     font-family: @webfont-italic, Arial, sans-serif;
     font-style: italic;
     font-weight: normal;
+
     .lt-ie9 & {
         font-style: normal !important;
     }
 }
 
-.webfont-medium() {
+.u-webfont-medium() {
     font-family: @webfont-medium, Arial, sans-serif;
     font-style: normal;
     font-weight: 500;
+
     .lt-ie9 & {
         font-weight: normal !important;
     }
 }
 
-.webfont-demi() {
+.u-webfont-demi() {
     font-family: @webfont-demi, Arial, sans-serif;
     font-style: normal;
     font-weight: bold;
+
     .lt-ie9 & {
         font-weight: normal !important;
     }
@@ -60,15 +63,16 @@
 //
 
 body {
+    .u-webfont-regular();
+
     color: @text;
-    font-family: Georgia, "Times New Roman", serif;
-    font-size: unit(@base-font-size-px / 16 * 100, %);
+    font-size: unit( @base-font-size-px / 16 * 100, % );
     line-height: @base-line-height;
 }
 
 .heading-1( @fs: @size-i ) {
     @font-size: @fs;
-    .webfont-regular();
+    .u-webfont-regular();
 
     margin-bottom: unit( 15px / @font-size, em );
     font-size: unit( @font-size / @base-font-size-px, em );
@@ -77,7 +81,7 @@ body {
 
 .heading-2( @fs: @size-ii ) {
     @font-size: @fs;
-    .webfont-regular();
+    .u-webfont-regular();
 
     margin-bottom: unit( 15px / @font-size, em );
     font-size: unit( @font-size / @base-font-size-px, em );
@@ -86,7 +90,7 @@ body {
 
 .heading-3( @fs: @size-iii )  {
     @font-size: @fs;
-    .webfont-regular();
+    .u-webfont-regular();
 
     margin-bottom: unit( 15px / @font-size, em );
     font-size: unit( @font-size / @base-font-size-px, em );
@@ -95,7 +99,7 @@ body {
 
 .heading-4( @fs: @size-iv )  {
     @font-size: @fs;
-    .webfont-medium();
+    .u-webfont-medium();
 
     margin-bottom: unit( 15px / @font-size, em );
     font-size: unit( @font-size / @base-font-size-px, em );
@@ -104,7 +108,7 @@ body {
 
 .heading-5( @fs: @size-v )  {
     @font-size: @fs;
-    .webfont-demi();
+    .u-webfont-demi();
 
     margin-bottom: unit( 15px / @font-size, em );
     font-size: unit( @font-size / @base-font-size-px, em );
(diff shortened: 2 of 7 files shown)


Scotchester:
Anyone know why the build might be failing because the tests can't find jQuery? I don't think it's related to anything I did here.

jimmynotjim:
@Scotchester I've seen the exact same thing happen in other PRs, and go away inexplicably. I'm thinking we ignore it for now and see how things end up when we're ready to merge this to canary.

package.json
```diff
@@ -25,7 +25,8 @@
     "postinstall": "node scripts/npm/postinstall",
     "release": "./scripts/travis/release.sh",
     "build": "gulp build",
-    "test": "gulp test && node test/accessibility/wcag.js"
+    "test": "gulp test && node test/accessibility/wcag.js",
+    "cf-link": "cd ./src/cf-core && npm link && cd ../cf-buttons/ && npm link"
```
jimmynotjim: This will definitely be unruly and is going to require some longer term thinking. Maybe we can hook into a manifest and loop these?

src/cf-core/src/cf-base.less
```diff
@@ -548,13 +538,13 @@ select[multiple] {
...
-    padding: unit(6 / @base-font-size-px, em);
-    font-family: Arial, sans-serif;
-    font-size: unit(@font-size / @base-font-size-px, em);
-    background: @input-bg;
+    padding: unit( 6 / @base-font-size-px, em );
     border: 1px solid @input-border;
     border-radius: 0;
+    margin: 0;
+    background: @input-bg;
+    font-family: Arial, sans-serif;
```
jimmynotjim: Should this be getting the `u-webfont-regular()` mixin? It won't be themed otherwise right?

src/cf-core/src/cf-utilities.less
```diff
@@ -246,38 +248,38 @@
...
+.u-small-text( @context: @base-font-size-px ) {
+    font-size: unit( 14px / @context, em );
 }
 
 .u-small-text {
-  .u-small-text();
+    .u-small-text();
 }
 
 small {
```
jimmynotjim: Since this is the exact same as `u-small-text` could we just combine them to a single comma delimited rule?

src/cf-core/src/cf-vars.less
```diff
@@ -32,9 +32,6 @@
 @bp-lg-max:                     1230px;
 @bp-xl-min:                     @bp-lg-max + 1px;
 
-// To be deprecated
```
jimmynotjim: Yay 👏

src/cf-core/src/cf-base.less
Scotchester: I think we intentionally wanted to avoid using the webfont in form fields? The Design Manual [doesn't explicitly say](http://cfpb.github.io/design-manual/page-components/form-fields.html), but the examples all show Arial.

Scotchester: Which, I suppose, isn't saying much, since it's using CF…

src/cf-core/src/cf-utilities.less
Scotchester: Good call. Done.

src/cf-core/src/cf-base.less
jimmynotjim: Hmmm, cfgov-refresh has the webfonts http://www.consumerfinance.gov/about-us/blog/

Scotchester: :shruggie:

ajbush:
Hey was talking to @jimmynotjim and for V1 we had been using Avenir for input fields. I'd vote to keep using that unless there's a technical reason why arial is better. It might also be worth checking with the team that worked on forms.

Scotchester:
@ajbush No technical reason. I think it's a _little_ bit strange for form fields not to use a system font, but I don't feel that strongly about it.

Is it something we should get a quick vote from designers on? And if we go that route, we should be sure to note it in the DM.

src/cf-core/src/cf-base.less
KimberlyMunoz: Do we need clarification here still? It might be worth asking in the Graphic Design manual what the intent was an either fixing here or opening in issue in cfgov-refresh.

src/cf-core/src/cf-base.less
```diff
@@ -548,13 +539,13 @@ select[multiple] {
     @font-size: 16px;
 
     display: inline-block;
-    margin: 0;
-    padding: unit(6 / @base-font-size-px, em);
-    font-family: Arial, sans-serif;
-    font-size: unit(@font-size / @base-font-size-px, em);
-    background: @input-bg;
+    padding: unit( 6 / @base-font-size-px, em );
```
KimberlyMunoz: Should this be `6px` instead of `6`?

src/cf-core/src/cf-utilities.less
```diff
@@ -33,14 +34,14 @@
...
-  overflow: hidden;
-  clip: rect(0 0 0 0);
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    border: 0;
+    margin: -1px;
+    padding: 0;
+    overflow: hidden;
+    clip: rect(0 0 0 0);
```
KimberlyMunoz: Should this spaced out with commas?

src/cf-core/src/cf-utilities.less
```diff
@@ -246,38 +238,36 @@
...
 // Small text utility
 //
 
-.u-small-text(@context: @base-font-size-px) {
-  font-size: unit(14px / @context, em);
+.u-small-text( @context: @base-font-size-px ) {
+    font-size: unit( 14px / @context, em );
 }
 
+small,
```
KimberlyMunoz: Nice.

src/cf-core/src/cf-vars.less
```diff
@@ -32,9 +34,6 @@
...
-@mobile-max:                    @bp-xs-max;
-@tablet-min:                    @bp-sm-min;
 
 // Color variables
 // $color- variables are from 18F's US Web Design Standards
```
KimberlyMunoz: U.S. Web Design Standards.

src/cf-core/usage.md
```diff
@@ -245,32 +242,30 @@ Adds a `.lt-ie8` fallback to hack inline block for IE 7 and below.
 
 Force word breaks within an element. Useful for small containers where text may over-run the width of the container.
```
KimberlyMunoz: Semantic line break?

src/cf-core/src/cf-media-queries.less
```diff
@@ -29,6 +29,7 @@
     }
 }
 
+// TODO: Discuss whether to split this into min and max queries.
 .respond-to-dpi(@ratio, @rules) {
```
KimberlyMunoz: Should this file follow the same spacing guidelines for variables in parenthesis that the other files do?

src/cf-core/src/cf-base.less
Scotchester: I just rearranged the line, but probably so!

src/cf-core/src/cf-utilities.less
Scotchester: Hmm… tricky one. [According to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/clip), this property is deprecated and should not be used. But [it's still present](https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css#L126) in the HTML5 Boilerplate (from which we originally cribbed this utility).

I'm not sure if removing it would break the utility, so I think we should leave it for now. The MDN page states that both space- and comma-separated syntaxes are separate, but space-separated is more backward-compatible.

src/cf-core/usage.md
Scotchester: I'm going to hold off on this to hopefully avoid conflicts with the standardized cf-core `usage.md` file that has been merged into `canary`

src/cf-core/src/cf-media-queries.less
Scotchester: Yes it should! Good catch.

src/cf-core/src/cf-base.less
Scotchester: @ajbush commented below. Still no clear answer.

Scotchester:
Fixes pushed. This should be good to go unless we get a definitive direction to change the `input` font. @ajbush, do you think it's something we need to check with other designers on before deciding to do it?

src/cf-core/src/cf-utilities.less
KimberlyMunoz: Then can you please add a comment that commas aren't being used for backwards-compatibility?

I know that screenreader-only utilities like these like to have their bases covered with more CSS than is necessary to make sure it doesn't render for sighted users, and I'm all in favor of that.

src/cf-core/usage.md
KimberlyMunoz: /shrug. We're going to have so many conflicts.

ajbush:
Hey @Scotchester I got feedback from a few designers that also thought it would be nice to use Avenir. I'd say let's use Avenir! Apologies for not replying sooner.

Scotchester:
Updated to set `input`s to use Avenir and add a requested comment.

Assigned to @CFPBot and ready for final review.

Scotchester:
(Travis still failing on unexplained missing jQuery issue.)

jimmynotjim:
👍  

I'm going to back merge `canary` into `prerelease-ver-4` once I get the last few doc PRs I started updated.

jimmynotjim:
👍

cfarm:
👍

PR #350, opened 2016-07-20
Convert Select to Atomic Design

cfarm:
Part of Atomic design issue https://github.com/cfpb/capital-framework/issues/345

Adds cf-forms from cfgov-refresh (based on @anselmbradford's update_form_elems branch) + makes select element a molecule

## Additions

- cf-forms.less from cfgov-refresh
- select element CSS from cfpb/cfgov-refresh#1273: New inputs and textareas
- markup for select elements in `usage.md`

## Testing

- npm link this and check out the following branch on capital-framework-sandbox to test: https://github.com/cfarm/capital-framework-sandbox/tree/atomic/cf-forms

## Review

- @jimmynotjim or @KimberlyMunoz 

## Screenshots
screen shot 2016-07-20 at 11 32 01 am
screen shot 2016-07-20 at 11 31 52 am

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #384, opened 2016-08-16
Big checkboxes

KimberlyMunoz:
This adds the big checkboxes to Capital Framework from the work done in https://github.com/cfpb/cfgov-refresh/pull/2292.

A lot of design decisions were made in the PR above and I did a little work improving/simplifying how we use variable names.

⚠️ I noticed that our labels are not up-to-spec. We don't have a task for labels and it's out of scope to handle here. I haven't adjusted those at all and will ask the Dada team when we should tackle this. 

## Testing

In this folder
```sh
cd cf-forms
npm link
```

Pull down this PR: https://github.com/jimmynotjim/capital-framework-sandbox/pull/18 and run `npm link cf-forms` then `npm start`

You should be able to navigate to the form section to see the new buttons. Check the boxes to see what they look like when active.

## Screenshots

👉 https://github.com/jimmynotjim/capital-framework-sandbox/pull/18

### DIFF
--- a/src/cf-forms/molecules/big-checkboxes.less
+++ b/src/cf-forms/molecules/big-checkboxes.less
@@ -1,0 +1,92 @@
+.m-big-checkbox {
+    margin-bottom: unit( 15px / @base-font-size-px, em );
+
+    &:last-child {
+        margin-bottom: unit( 30px / @base-font-size-px, em );
+    }
+    .a-input {
+        @checkbox-width: unit( 18px / @base-font-size-px, em );
+
+        &[type="checkbox"] {
+            .u-visually-hidden();
+
+            .lt-ie9 & {
+                height: 20px;
+                width: 20px;
+                border: 0;
+                float: left;
+                margin: 1em;
+                position: static;
+                width: auto;
+            }
+        }
+
+        &[type="checkbox"] + label {
+            cursor: pointer;
+            margin-bottom: 0;
+        }
+
+        &[type="checkbox"] + label::before {
+            background: white;
+            content: '\a0';
+            display: inline-block;
+            margin-top: unit( 2px / @base-font-size-px, em );
+            margin-right: 0.6em;
+        }
+
+        &[type="checkbox"] + label::before {
+            height: @checkbox-width;
+            width: @checkbox-width;
+            box-shadow: 0 0 0 1px @input-bg, 0 0 0 2px @input-border;
+        }
+
+        &[type="checkbox"]:checked + label::before {
+            content: @cf-icon-approved;
+            font-family: 'CFPB Minicons';
+            font-size: unit( 16px / @base-font-size-px, em );
+            box-shadow: 0 0 0 1px @input-bg, 0 0 0 2px @input-active;
+            text-align: center;
+        }
+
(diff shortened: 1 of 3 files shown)


src/cf-forms/molecules/big-checkboxes.less
```diff
@@ -0,0 +1,89 @@
...
+    &:last-child {
+        margin-bottom: unit( 30px / @base-font-size-px, em );
+    }
+    .a-input {
+        @checkbox-width: unit( 18px / @base-font-size-px, em );
+
+        &[type="checkbox"] {
+            .u-visually-hidden();
+
+            .lt-ie9 & {
```
cfarm: properties should be ordered according to front-end standards:

https://github.com/cfpb/front-end/blob/master/css.md#property-order

src/cf-forms/src/cf-forms.less
```diff
@@ -30,12 +30,27 @@
 // .disabled
 @input-disabled:                #aeb0b5; // $color-gray-light
 
+// Large target area variables
+@input-big-target-bg:           #d6d7d9; // $color-gray-lighter
```
cfarm: what do these varaibles in the comments refer to? US Web Design Standards?

src/cf-forms/src/cf-forms.less
```diff
@@ -30,12 +30,27 @@
...
+
+@input-big-target-bg__disabled: #aeb0b5; // $color-gray-light
+
 // Sizing variables
 
 // .input__super
 @input__super-font-size:        18px;
 
 //
+// Import molecules and atoms
```
cfarm: Can this go up top, before the color variables and after the other imports? I would expect to find all the imports in one place.

cfarm:
I'm done reviewing - One comment that needs a commit, one question and one suggested change.

src/cf-forms/src/cf-forms.less
cfarm: Also I like this structure, should I use the same for the select in PR #350? (in a subsequent pr, that one has languished long enough)

src/cf-forms/src/cf-forms.less
KimberlyMunoz: Yep. It's mentioned in L:19.

src/cf-forms/src/cf-forms.less
KimberlyMunoz: We'd get an error if we moved them before the color variables. I could move the other imports down.

src/cf-forms/molecules/big-checkboxes.less
KimberlyMunoz: Whoops, I thought positioning went before display.

src/cf-forms/src/cf-forms.less
cfarm: Ah, ok. Yes, let's keep them together.

KimberlyMunoz:
Thanks, I've made some changes based on your reviews in 1849bdb

cfarm:
👍

cfarm:
AndConf inspired bigfoot gif

bigfoot shaking booty

PR #385, opened 2016-08-18
Big radio buttons

KimberlyMunoz:
Big Radio Buttons. The other half of #384 

⚠️ I noticed that our labels are not up-to-spec. We don't have a task for labels and it's out of scope to handle here. I haven't adjusted those at all and will ask the Dada team when we should tackle this. 

## Testing

In this folder
```sh
cd src/cf-forms
npm link
```

Pull down this PR: https://github.com/jimmynotjim/capital-framework-sandbox/pull/19 and run `npm link cf-forms` then `npm start`

You should be able to navigate to the form section to see the new buttons. Check the buttons to see what they look like when active.

## Screenshots

👉 https://github.com/jimmynotjim/capital-framework-sandbox/pull/19

### DIFF
--- a/src/cf-forms/molecules/big-radio-buttons.less
+++ b/src/cf-forms/molecules/big-radio-buttons.less
@@ -1,0 +1,87 @@
+.m-big-radio-button {
+    margin-bottom: unit( 15px / @base-font-size-px, em );
+
+    &:last-child {
+        margin-bottom: unit( 30px / @base-font-size-px, em );
+    }
+    .a-input {
+        @radio-width: unit( 17px / @base-font-size-px, em );
+
+        &[type="radio"] {
+            .u-visually-hidden();
+
+            .lt-ie9 & {
+                height: 20px;
+                width: 20px;
+                border: 0;
+                float: left;
+                margin: 1em;
+                position: static;
+                width: auto;
+            }
+        }
+
+        &[type="radio"] + label {
+            cursor: pointer;
+            margin-bottom: 0;
+        }
+
+        &[type="radio"] + label::before {
+            display: inline-block;
+            margin-top: unit( 2px / @base-font-size-px, em );
+            margin-right: 0.6em;
+            border-radius: 100%;
+            box-shadow: 0 0 0 2px #fff, 0 0 0 3px @input-border;
+            height: @radio-width;
+            width: @radio-width;
+            background: white;
+            content: '\a0';
+        }
+
+        &[type="radio"]:checked + label::before {
+            background-color: @input-border__selected;
+            box-shadow: 0 0 0 2px @input-bg, 0 0 0 3px @input-border__selected;
+        }
+
+        &[type="radio"]:hover + label::before {
+            box-shadow: 0 0 0 2px @input-bg, 0 0 0 3px @input-border__selected;
+        }
+
+        &[type="radio"]:focus + label::before {
+            box-shadow: 0 0 0 1px @input-bg, 0 0 0 3px @input-border__selected;
+        }
+
+        &[type="radio"]:disabled + label {
+            color: @input-disabled;
+        }
+
+        &[type="radio"]:disabled + label::before {
+            background: @input-big-target-bg__disabled;
+            box-shadow: 0 0 0 1px @input-bg, 0 0 0 2px @input-disabled;
+            cursor: not-allowed;
+        }
+    }
+
+    label {
+        margin: 0;
+        padding: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em )
+                 unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em )
+                 unit( ( @grid_gutter-width / 1.75 ) / @base-font-size-px, em );
+    }
+
+    &.selected label,
+    &.focus label,
+    input:focus + label,
+    input:checked + label {
+        outline: 1px solid @input-border-focus;
+    }
+
+    &.selected label,
+    input:checked + label {
+        background-color: @input-big-target-bg__selected;
+    }
+
+    &_container {
+        background: @input-big-target-bg;
+    }
+}
--- a/src/cf-forms/molecules/big-radio-buttons.less
+++ b/src/cf-forms/molecules/big-radio-buttons.less
@@ -14,10 +14,9 @@
                 height: 20px;
                 width: 20px;
                 border: 0;
-                float: left;
                 margin: 1em;
                 position: static;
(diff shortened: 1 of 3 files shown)


jimmynotjim:
Should disabled checkboxes and radios get the same cursor: not-allowed we use on disabled buttons?

src/cf-forms/molecules/big-radio-buttons.less
```diff
@@ -0,0 +1,86 @@
+.m-big-radio-button {
+    margin-bottom: unit( 15px / @base-font-size-px, em );
```
jimmynotjim: These two margin bottom styles should go in the container organism right?

src/cf-forms/molecules/big-radio-buttons.less
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        &[type="radio"]:disabled + label {
+            color: @input-disabled;
+        }
+
+        &[type="radio"]:disabled + label::before {
+            background: @input-big-target-bg__disabled;
+            box-shadow: 0 0 0 1px @input-bg, 0 0 0 2px @input-disabled;
+            cursor: not-allowed;
```
jimmynotjim: Ah I see now, this needs to be on the label instead of the psuedo-elem because you can still attempt to click the big box around the label.

src/cf-forms/molecules/big-radio-buttons.less
```diff
@@ -0,0 +1,86 @@
...
+    label {
+        padding: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em )
+                 unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em )
+                 unit( ( @grid_gutter-width / 1.75 ) / @base-font-size-px, em );
+        margin: 0;
+    }
+
+    &.selected label,
+    &.focus label,
+    input:focus + label,
```
jimmynotjim: Should this be `.a-input:focus + label` to stay consistent w/ the code above?

src/cf-forms/molecules/big-radio-buttons.less
```diff
@@ -0,0 +1,86 @@
+.m-big-radio-button {
+    margin-bottom: unit( 15px / @base-font-size-px, em );
+
+    &:last-child {
+        margin-bottom: unit( 30px / @base-font-size-px, em );
+    }
+    .a-input {
+        @radio-width: unit( 17px / @base-font-size-px, em );
```
jimmynotjim: Variables have typically been declared at the top of the file for easy reference.

src/cf-forms/molecules/big-radio-buttons.less
KimberlyMunoz: I'd prefer to keep it as is for ease of use. Without this code, the buttons don't work period.

src/cf-forms/molecules/big-radio-buttons.less
KimberlyMunoz: I don't see. Can you clarify your feedback?

src/cf-forms/molecules/big-radio-buttons.less
KimberlyMunoz: What are the advantages of that versus leaving it as, especially when this parent class already has margins from the grid mixin. Wouldn't it make sense to keep them together when inspecting?

src/cf-forms/molecules/big-radio-buttons.less
jimmynotjim: I can't take a screenshot of the mouse hand, but hover the gray box of a disabled big radio and the hand appears instead of the not-allowed cursor you get when you hover the psudeo-elem.

src/cf-forms/molecules/big-radio-buttons.less
jimmynotjim: Weird, maybe leave a note so no one attempts to change it later.

src/cf-forms/molecules/big-radio-buttons.less
KimberlyMunoz: We've had a rule that if it's only being used in one class, it can be on top of that class, but this sizing variable will probably be used if we move classic checkboxes/radios over. Should I move to cf-forms or are we still expecting to not move those over?

src/cf-forms/molecules/big-radio-buttons.less
KimberlyMunoz: 2ef1fbd?

src/cf-forms/molecules/big-radio-buttons.less
KimberlyMunoz: I mean, it's a stylistic choice vs being really explicit about the class `a-input`. One I prefer so the buttons continue to work even if `.a-input` isn't included in the large checkbox. I don't think it's something at risk for removal, unless you have another use case in mind?

src/cf-forms/molecules/big-radio-buttons.less
jimmynotjim: I can't find the conversation now, but @anselmbradford did a good job explaining it at one point. The gist was that a elements should not "know" anything about it's surroundings, including spacing against other elements and the next parent should be used to set those instead.

KimberlyMunoz: Well, if we go down that path, we'd have to use nth of type and that wouldn't do the correct spacing when there's a form with more than one checkbox or radio button, plus all the grid stuff would have to be refactored again.

src/cf-forms/molecules/big-radio-buttons.less
jimmynotjim: No, it's just something I'd probably standardize later if I didn't know (or forgot) this conversation happened. If it breaks things, we should make it clear to folks it can't be changed.

src/cf-forms/molecules/big-radio-buttons.less
jimmynotjim: Yep, looks like that did it.

src/cf-forms/molecules/big-radio-buttons.less
jimmynotjim: Lets leave it then and we can hash out spacing later.

src/cf-forms/molecules/big-radio-buttons.less
jimmynotjim: Leave it for now, it can get moved when that work happens.

jimmynotjim:
👍

PR #389, opened 2016-08-22
Fix disabled select state hover

cfarm:
Small fix to remove the blue outline on hover from the disabled select element.

## Additions

- hover style for disabled select to return the outline to its default transparent style.

## Review

- @jimmynotjim 

#### Before
screen shot 2016-08-22 at 6 26 56 pm

#### After
screen shot 2016-08-22 at 6 26 43 pm

## Notes

- focus and active styles are irrelevant here since there is no focus or active state on a disabled input (browser default behavior)
- hover style is the same as the default select outline style

## Checklist

### DIFF
--- a/src/cf-forms/src/cf-forms.less
+++ b/src/cf-forms/src/cf-forms.less
@@ -343,6 +343,10 @@ textarea {
     select[disabled] {
         color: @input-disabled-text;
         background-color: @input-disabled;
+
+        &:hover {
+            outline: 2px solid transparent;
+        }
     }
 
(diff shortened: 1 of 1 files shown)


jimmynotjim:
One other thing I noticed while testing the big buttons is we should probably add `cursor: not-allowed` to these disabled selects as well.

cfarm:
is that not default browser behavior? hmmmm. I'll make a new issue for it

jimmynotjim:
👍

PR #390, opened 2016-08-25
Remove .form-group

KimberlyMunoz:
Form group is currently being used inconsistently across many projects
for layout purposes. This commit removes it, we'll have a PR for how we
can handle atomic design and form layout this sprint.

## Review
- @cfarm


jimmynotjim:
👍

PR #393, opened 2016-08-29
Moved base form styles from cf-core to cf-forms

jimmynotjim:
This doesn't do anything other than move the styles so they're organized in one place. It's a first step in a larger atomic re-write of our form styles. This change is bound to be contentious and therefore deemed important enough to do separate from everything else.

## Changes

- Moved base form styles to cf-forms

## Testing

- checkout and run `npm run cf-link`
- open new tab and cd to your sandbox repo
- in the sandbox tab run `npm link cf-forms`
- in the sandbox tab run `npm run build`
- in the sandbox tab run `npm start`
- navigate to the cf-forms examples and ensure everything looks as it did (mostly inputs)

## Review

- @mistergone 
- @cfarm 
- @Scotchester 

## Notes

- @Scotchester has voiced that he's not in favor of this change because they are basic html elements not needing classes. While I agree, I find it's super confusing that form elements are the only ones styled in two places and believe this is a usability improvement for new (and even existing) developers.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 - **cf-layout:** [MAJOR] Updated featured content module to atomic naming conventions
   - added featured content module modifiers for right and center image anchors
 - **cf-pagination** [MAJOR] Update to atomic naming conventions:
+- **cf-core** [MAJOR] Moved the base form styles to cf-forms
 
 ### Removed
 - **cf-core:** [MAJOR] Removed deprecated items:
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -515,87 +515,8 @@ blockquote {
 }
 
 //
-// Form labels
+// Form elements have been moved to the cf-forms component
 //
-
-label {
-    display: block;
-    // 5px is eyeballed to 10px when considering line height.
-    margin-bottom: unit( 5px / @base-font-size-px, em );
-}
-
-//
-// Form elements
-//
-
-input[type="text"],
-input[type="search"],
-input[type="email"],
-input[type="url"],
-input[type="tel"],
-input[type="number"],
-textarea,
-select[multiple] {
-    @font-size: 16px;
-    .u-webfont-regular();
-
-    display: inline-block;
-    padding: unit( 6px / @base-font-size-px, em );
-    border: 1px solid @input-border;
-    border-radius: 0;
-    margin: 0;
-    background: @input-bg;
-    font-size: unit( @font-size / @base-font-size-px, em );
-    vertical-align: top;
-    appearance: none;
-}
-
-// Overrides extra left padding.
-// http://stackoverflow.com/questions/11127891/how-can-i-get-rid-of-horizontal-padding-or-indent-in-html5-search-inputs-in-webk
-::-webkit-search-decoration {
-    appearance: none;
-}
-
-// The .focus class is only included for documentation demos and should not
-// be used in production.
-input[type="text"]:focus,
-input[type="text"].focus,
-input[type="search"]:focus,
-input[type="search"].focus,
-input[type="email"]:focus,
-input[type="email"].focus,
-input[type="url"]:focus,
-input[type="url"].focus,
-input[type="tel"]:focus,
-input[type="tel"].focus,
-input[type="number"]:focus,
-input[type="number"].focus,
-textarea:focus,
-textarea.focus,
-select[multiple]:focus,
-select[multiple].focus {
-    border: 1px solid @input-border-focus;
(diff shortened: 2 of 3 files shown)


jimmynotjim:
Here's a preview of why this will make a difference. With base styles and cf-input combined, we end up with 

```
.a-text-input {
    .u-webfont-regular();

    // Browser Reset
    appearance: none;

    display: inline-block;

    box-sizing: border-box;
    padding: unit( 6px / @base-font-size-px, em );
    width: 100%;

    background: @input-bg;
    vertical-align: top;

    &:focus,
    &__focused {
        border-color: @input-border-focus;
        outline: 1px solid @input-border-focus;
    }
}
```

Which will allows us to easily extend the styles to other states and situations, instead of having to use classes in some places and element selectors in others ex: https://github.com/cfpb/capital-framework/pull/385#discussion_r75884973

mistergone:
Looks good to me!

colber-thumbs-up-wink

Scotchester:
It seems like this is about two distinct issues:

1. Going to work on cf-forms and not knowing that base styles are set in cf-core.
   - It seems like this could easily be solved with a prominent code comment notifying the user of that fact.
   - It's not really true that this is the only place this happens. For example, basic list styling happens in core, but fancy list stuff happens in cf-typography.
2. Wanting to have a class on inputs for the purpose of … well, I'm not entirely sure.
   - > Which will allows us to easily extend the styles to other states and situations

     What possible other situations would we use this styling for?
   - > instead of having to use classes in some places and element selectors

     What's wrong with that? It already happens with stuff like lists. It's simple CSS. Are there places where it's causing actual cascade/specificity/overriding issues that I"m just not seeing? I don't totally understand the discussion that you referenced in https://github.com/cfpb/capital-framework/pull/385#discussion_r75884973
   - I still don't like having to use a class for _every usage_ of a basic HTML element. If I'm never going to use the element _without_ that class, why force myself to remember to add the class every time, or else get a default-browser-styled element?

So, currently still a :-1: from me. But open to being convinced.

jimmynotjim:
> It seems like this could easily be solved with a prominent code comment notifying the user of that fact.

But why? Why change it in two places when it's far easier in one? What's the selling point besides "we've always done it like this"? As far as I've seen we don't ever use cf-core without using cf-forms, so I don't see the value in keeping the separated.

> It's not really true that this is the only place this happens. For example, basic list styling happens in core, but fancy list stuff happens in cf-typography.

Lists a much different use case since they are often used without any attributes whereas inputs will always have attributes because they require a name, id, and type as well as often having a placeholder and value depending on the type.

> What possible other situations would we use this styling for?

There are already three instances for inputs, error, warning, and success. To keep with the atomic principles, the first three should be `.name__state` instead of `.name.state`. That's not possible and a lot messier to do with a bunch of `elem[attr]` selectors.

What's wrong with that? It already happens with stuff like lists. Where? There are no attr selectors in lists as far as I know so there are no specificity issues when trying to use a class instead.

> If I'm never going to use the element without that class

See above, inputs are never used without markup attributes. Adding a class isn't a big deal and it's probably automated 9 times out of 10 with templates.

Scotchester:
> But why? Why change it in two places when it's far easier in one? What's the selling point besides "we've always done it like this"? As far as I've seen we don't ever use cf-core without using cf-forms, so I don't see the value in keeping them separated.

I said "comment" not "duplicated code" :) There would be no changing things in multiple places. The selling point is that is just seems natural _(to me)_ to do basic element styling all in the same place.

> Lists are a much different use case since they are often used without any attributes whereas inputs will always have attributes because they require a name, id, and type as well as often having a placeholder and value depending on the type.

I don't see how this is relevant to where the Less code lives, which is what I was talking about when making that particular point.

> > What possible other situations would we use this styling for?
> There are already three instances for inputs, error, warning, and success. To keep with the atomic principles, the selectors should be .name__state instead of .name.state or elem.state. That's not possible and a lot messier to do with a bunch of elem[attr] selectors.
>
>  > What's wrong with that? It already happens with stuff like lists.
>
> Where? There are no attr selectors in lists as far as I know so there are no specificity issues when trying to use a class instead.

And finally we come to the point where I begin to understand that **attribute selectors are the actual problem, causing specificity issues**. I wasn't making that connection, and it was never stated explicitly. In fact, attribute selectors have not been mentioned at all prior to your last post :)

I don't like that we have to do it, and if it were just a personal project, I would deal with the specificity issues to conform to my vision of the "ideal" (in terms of code organization and not requiring classes on basic elements), but I get it now. :+1: for making it somewhat simpler for a team dealing with it.

jimmynotjim:
Will do. 

As for the two places thing, I didn't mean make the same changes in two places, I meant if you were adjusting the inputs later in life, and it turned out a base adjustment also messed with cf-forms, you're now making changes in two places.

PR #398, opened 2016-09-07
Adding super input

KimberlyMunoz:
Adds the super input atom to cf-forms. 

## Testing

Test with this branch right here: https://github.com/cfpb/capital-framework-sandbox/pull/25


KimberlyMunoz:
As I made the sandbox branch, I saw they weren't really aligned. Comparing to see how I can fix.

image

src/cf-forms/src/cf-forms.less
```diff
@@ -2,6 +2,13 @@
    Capital Framework
    Form Element Styling
    ========================================================================== */
+// Import external dependencies
```
KimberlyMunoz: I know we moved this to keep things together, but if we have variables that depend on the dependencies, we need to import them before the variables.

PR #399, opened 2016-09-09
Remove the super input

KimberlyMunoz:
Look at that deleted code. Looks good right? As @jimmynotjim and I were pairing on #398, we realized that the super input isn't in the design manual nor being used. Since the atomic conversion is a good time to remove stuff we aren't using, this might be a great candidate for removal.

## Removals

- `input__super`

## Review

- @cfpb/cfgov-frontends 

## Notes

- Obviously, we either merge #398 or this one, not both 😉


jimmynotjim:
👏  but since I've already weighed in while pairing, I'll let others sign off if they agree

Scotchester:
:+1:

PR #403, opened 2016-09-14
Add atomic inline form validation

KimberlyMunoz:
Removed the older form validation and replaced with atomic form
validation following the guidelines here:
http://cfarm.github.io/design-manual/ui-toolkit/form-fields.html

## Testing

- Pull down this branch to test on another repo. https://github.com/cfpb/capital-framework-sandbox/pull/28

## Review

- @cfpb/cfgov-frontends 

## Screenshots

image

### DIFF
--- a/src/cf-forms/src/molecules/field.less
+++ b/src/cf-forms/src/molecules/field.less
@@ -104,7 +104,7 @@
 
     &__error {
         .a-error-message {
-            margin-top: 15px;
+            margin-top: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
         }
     }
 }
--- a/src/cf-forms/src/molecules/field.less
+++ b/src/cf-forms/src/molecules/field.less
@@ -101,4 +101,14 @@
             }
         }
     }
+
+    &__error {
+        .a-error-message {
+            margin-top: 15px;
+        }
+    }
+}
+
+.a-error-message .cf-icon-delete-round {
+    color: @input-icon__error;
 }
--- a/src/cf-forms/usage.md
+++ b/src/cf-forms/usage.md
@@ -100,30 +100,6 @@ Inputs should always be paired with a label for accessibility reasons.
 
 See the 'Form icons' section below for guidance on adding icons to states.
 
-#### Error state
-
-<input class="error" type="text" value="Invalid input" title="Test input">
-
-```
-<input class="error" type="text" value="Invalid input" title="Test input">
-```
-
-#### Warning state
-
-<input class="warning" type="text" value="Invalid input" title="Test input">
-
-```
-<input class="warning" type="text" value="Invalid input" title="Test input">
-```
-
-#### Success state
-
-<input class="success" type="text" value="Validated input" title="Test input">
-
-```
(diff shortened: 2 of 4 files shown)


src/cf-forms/usage.md
```diff
@@ -110,16 +86,34 @@ See the 'Form icons' section below for guidance on adding icons to states.
...
-<input type="text" value="" title="Test input">
-<span class="cf-form_input-icon cf-icon cf-icon-email"></span>
+<div class="m-field m-field__error">
+    <label class="a-label__heading" for="form-input-error">Label</label>
+    <input class="a-text-input a-text-input__error"
+           type="text"
+           value="Invalid input"
+           id="form-input-error"
+           aria-describedby="form-input-error_message">
+    <div class="a-error-message" id="form-input-error_message" role="alert">
```
jimmynotjim: What do you think about calling this `a-field-alert a-field-alert__error` so that it's extensible? That would also line it up with the naming in the Design Manual fork you've referenced.

src/cf-forms/usage.md
```diff
@@ -110,16 +86,34 @@ See the 'Form icons' section below for guidance on adding icons to states.
...
+        <span class="cf-icon cf-icon-delete-round" aria-hidden="true"></span>
```
jimmynotjim: Do we need the span? If it was a modifier as suggested above we could use a pseudo-elem and update it based on the alert type.

src/cf-forms/src/molecules/field.less
```diff
@@ -101,4 +101,14 @@
             }
         }
     }
+
+    &__error {
+        .a-error-message {
+            margin-top: 15px;
```
jimmynotjim: Should this be updated to the gutter variable divided by 2?

src/cf-forms/usage.md
KimberlyMunoz: If the only message is an error message, I think `a-error-message` is a little more compact. (This is also how the USWDS do it.) Are there other inline messages we need to consider that should be included in the whole framework and not a page-by-page rule?

src/cf-forms/src/molecules/field.less
KimberlyMunoz: Good call. Will change.

src/cf-forms/usage.md
jimmynotjim: I thought when we did this in v1 there was a warning state as well, but I'm not finding it now.

src/cf-forms/usage.md
KimberlyMunoz: I prefer the span with the already written rules than the pseudo-elem with more CSS to write. I'm slightly worried about the size and number of rules we have already.

src/cf-forms/src/molecules/field.less
KimberlyMunoz: Changed in a2a12cd

src/cf-forms/usage.md
jimmynotjim: I am as well, which is why I was in favor for a base class that gets modified vs multiple individual states, but if there are no other states, it's moot and this can all stay.

PR #405, opened 2016-09-16
Standardized Select with other form fields

KimberlyMunoz:
Standardized the select box to match the other form fields. Also renamed `.m-field` to `.m-form-field` to make it more clear where all the form fields live and differentiate between something like `fieldset`

## Testing

- Test against https://github.com/cfpb/capital-framework-sandbox/pull/27

## Screenshots

image

### DIFF
--- a/src/cf-forms/src/atoms/select.less
+++ b/src/cf-forms/src/atoms/select.less
@@ -1,0 +1,68 @@
+.a-select {
+    position: relative;
+    border: 1px solid @select-border;
+
+    select {
+        height: unit( 30px / @base-font-size-px, em );
+        width: 100%;
+        padding: unit( 4px / @base-font-size-px, em )
+                 unit( 6px / @base-font-size-px, em );
+        border: 0;
+        outline: 2px solid transparent;
+        appearance: none;
+        background-color: @input-bg;
+        border-radius: 0;
+        color: @text;
+
+        &:hover,
+        &:active,
+        &:focus {
+            outline-color: @input-border__focused;
+            outline-offset: 0;
+        }
+    }
+
+    select[disabled] {
+        color: @input-text__disabled;
+        background-color: @input-bg__disabled;
+        cursor: not-allowed;
+
+        &:hover {
+            outline: 2px solid transparent;
+        }
+    }
+
+    select option:disabled {
+        color: @input-disabled;
+    }
+
+    &:after {
+        width: unit( 30px / @base-font-size-px, em );
+        height: unit( 30px / @base-font-size-px, em );
+        border-left: 1px solid @select-border;
+        position: absolute;
+        top: 0;
+        right: 0;
+        background-color: @select-icon-bg;
+        color: @select-icon;
+        content: @cf-icon-down;
+        font-family: 'CFPB Minicons';
+        line-height: unit( 30px / @base-font-size-px, em );
+        text-align: center;
+        pointer-events: none;
+    }
+}
+
+
+// TODO Add modernizr to CF so this works
+// Right now modern IE doesn't support pointer events causing nothing
+// to happen when you click on the dropdown error.
+
+.no-csspointerevents .a-select {
+    &:after {
+        height: 0;
+        width: 0;
+        border: 0;
+        content: '';
+    }
+}
--- a/src/cf-forms/src/atoms/select.less
+++ b/src/cf-forms/src/atoms/select.less
@@ -28,7 +28,7 @@
         cursor: not-allowed;
 
         &:hover {
-            outline: 2px solid transparent;
+            outline: none;
         }
     }
 
--- a/src/cf-forms/src/atoms/select.less
+++ b/src/cf-forms/src/atoms/select.less
@@ -1,14 +1,14 @@
 .a-select {
+
     position: relative;
     border: 1px solid @select-border;
 
     select {
-        height: unit( 30px / @base-font-size-px, em );
+        height: unit( @select-height / @base-font-size-px, em );
         width: 100%;
         padding: unit( 4px / @base-font-size-px, em )
                  unit( 6px / @base-font-size-px, em );
         border: 0;
-        outline: 2px solid transparent;
         appearance: none;
         background-color: @input-bg;
         border-radius: 0;
@@ -17,7 +17,7 @@
         &:hover,
         &:active,
         &:focus {
-            outline-color: @input-border__focused;
+            outline: 2px solid @input-border__focused;
             outline-offset: 0;
         }
     }
@@ -37,8 +37,7 @@
     }
 
     &:after {
-        width: unit( 30px / @base-font-size-px, em );
-        height: unit( 30px / @base-font-size-px, em );
+        width: unit( @select-height / @base-font-size-px, em );
         border-left: 1px solid @select-border;
         position: absolute;
         top: 0;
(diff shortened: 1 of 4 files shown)


src/cf-forms/usage.md
```diff
@@ -96,6 +96,34 @@ Inputs should always be paired with a label for accessibility reasons.
 <textarea class="a-text-input" id="textarea-example">Lorem Ipsum</textarea>
 ```
 
+### Basic checkboxes
+
+<div class="m-form-field__checkbox">
+    <input class="a-checkbox" type="checkbox" id="test_checkbox">
+    <label class="a-label" for="test_checkbox">Label</label>
+</div>
+
```
KimberlyMunoz: Whoops. I forgot to add the code comments. Coming up shortly!

KimberlyMunoz: Fixed.

src/cf-forms/src/molecules/form-fields.less
```diff
@@ -101,4 +101,73 @@
             }
         }
     }
+
+    &__select {
+        .a-select {
+            position: relative;
+            margin-top: 0.5em;
```
jimmynotjim: Should everything other than the margin and width below be in `/atoms/select.less` since they're more about visual style than layout and they aren't combining multiple elements? I went back and forth on the label above because they're visual changes but eventually settled on this file because they were dependent on the state of the sibling elem. Everything here looks like it can exist entirely on it's own.

jimmynotjim: Also, the margin top should probably be done the same as on inputs, because we wouldn't want it added if there was no label inside the field, like if there's a label heading outside the field. That margin would be handled by the label heading itself.

src/cf-forms/src/molecules/form-fields.less
```diff
@@ -101,4 +101,73 @@
...
+                line-height: unit( 30px / @base-font-size-px, em );
+                text-align: center;
+                pointer-events: none;
+            }
+        }
+    }
+
+    // @todo: add modernizr to CF so this works:
+
+    .no-csspointerevents .m-select {
```
jimmynotjim: This needs to be updated to `.a-select`

src/cf-forms/src/molecules/form-fields.less
```diff
@@ -101,4 +101,73 @@
...
+            }
+
+            &:after {
+                width: unit( 30px / @base-font-size-px, em );
+                height: unit( 30px / @base-font-size-px, em );
+                border-left: 1px solid @select-border;
+                position: absolute;
+                top: 0;
+                right: 0;
+                background-color: @input-disabled;
```
jimmynotjim: This variable should be updated since this isn't a disabled input even if they're the same color.

src/cf-forms/src/molecules/form-fields.less
```diff
@@ -101,4 +101,73 @@
...
+                }
+            }
+
+            select[disabled] {
+                color: @input-text__disabled;
+                background-color: @input-bg__disabled;
+                cursor: not-allowed;
+
+                &:hover {
+                    outline: 2px solid transparent;
```
jimmynotjim: Why transparent? Could it be `none`?

KimberlyMunoz: That's not an option with outline and you want there to be that offset so that the select doesn't "jump" on hover.

src/cf-forms/src/molecules/form-fields.less
KimberlyMunoz: I'm a little conflicted on this one. Ideally, it would make sense to group all the form elements together, but I think separating it out should lead to shorter selectors, so I'll make the change.

KimberlyMunoz: Fixed in 45aa396

src/cf-forms/src/molecules/form-fields.less
KimberlyMunoz: Fixed in 45aa396

src/cf-forms/src/molecules/form-fields.less
KimberlyMunoz: Good catch! Fixed in 45aa396.

src/cf-forms/src/molecules/form-fields.less
jimmynotjim: We [use it in cf-tables](https://github.com/cfpb/capital-framework/blob/8b2807216061f88f5930040482f4c2fcfa1951a4/src/cf-tables/src/cf-tables.less#L54) and outline shouldn't be moving elements, it doesn't affect document flow. From https://developer.mozilla.org/en-US/docs/Web/CSS/outline

> Outlines do not take up space, they are drawn above the content.

KimberlyMunoz: Oooh. Tested and the change looks good, thanks. 085f019

src/cf-forms/src/atoms/select.less
```diff
@@ -0,0 +1,68 @@
...
+    position: relative;
+    border: 1px solid @select-border;
+
+    select {
+        height: unit( 30px / @base-font-size-px, em );
+        width: 100%;
+        padding: unit( 4px / @base-font-size-px, em )
+                 unit( 6px / @base-font-size-px, em );
+        border: 0;
+        outline: 2px solid transparent;
```
jimmynotjim: Is this needed? Since it doesn't affect flow I think you can limit it to hover and focus.

src/cf-forms/src/atoms/select.less
```diff
@@ -0,0 +1,68 @@
...
+        appearance: none;
+        background-color: @input-bg;
+        border-radius: 0;
+        color: @text;
+
+        &:hover,
+        &:active,
+        &:focus {
+            outline-color: @input-border__focused;
+            outline-offset: 0;
```
jimmynotjim: I've only seen this used on the button links. Do we need it if it's 0? It looks like [that is the default](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset).

src/cf-forms/src/atoms/select.less
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    select option:disabled {
+        color: @input-disabled;
+    }
+
+    &:after {
+        width: unit( 30px / @base-font-size-px, em );
+        height: unit( 30px / @base-font-size-px, em );
```
jimmynotjim: Should height be omitted and `bottom: 0` be used instead in case the sibling select changes in height?

src/cf-forms/src/atoms/select.less
KimberlyMunoz: This happens without the rule. 
image

src/cf-forms/src/atoms/select.less
KimberlyMunoz: Line-height would still need to change, so I've gone ahead and removed this rule since line-height is really setting the height here. d0c16c4

src/cf-forms/src/atoms/select.less
KimberlyMunoz: Removed in d0c16c4

src/cf-forms/src/atoms/select.less
jimmynotjim: weird. Never mind then.

jimmynotjim:
👍  Great work, thanks for checking on those two thoughts.

PR #412, opened 2016-10-11
Atomic/typography

jimmynotjim:
Converted cf-typography to atomic standards

## Removals

- Removed custom bullet mixin because it is no longer used
- Removed fancy slug because it is no longer used
- Removed short-desc because we are now using Avenir as our standard body font
- Removed padded heading because it's no longer used.
- Removed icon lists for the time being to be addressed later when we can look at existing implementations.

## Changes

- Broke links apart to create less repetitive code
- Broke headings and headers apart to create a better parent/child relationship and handle layout according to our best practices.

## Testing

- check out `example/cf-typography ` sandbox branch and run `npm run build && npm start`
- navigate to `http://localhost:3000/components/cf-typography/`

## Review

- @Scotchester 

## Screenshots

Too many to list, but everything should really look the same as the examples at `https://cfpb.github.io/capital-framework/components/cf-typography/#pull-quote` 

## Notes

- `a-heading` is a placeholder much like `a-list`. There are no uniform styles between the headings, but it makes the most sense, especially when controlling the layout/margin of stylized headings nested inside headers.
- __Reduces output by 4kb__. Small gains but still a positive impact.

## Todos

- Clean up the way we're handling icons. There's a ton of repetitive code due to icons and I have an idea of how it could be better implemented.

## Checklist

### DIFF
--- a/src/cf-typography/src/atoms/date.less
+++ b/src/cf-typography/src/atoms/date.less
@@ -1,0 +1,6 @@
+.a-date {
+    .heading-5();
+
+    color: @date;
+    white-space: nowrap;
+}
--- a/src/cf-typography/src/atoms/headings.less
+++ b/src/cf-typography/src/atoms/headings.less
@@ -1,0 +1,15 @@
+.a-heading__icon {
+    .heading-4();
+
+    color: @heading__icon;
+
+    a& {
+        .u-link__colors( @heading__icon, @heading__icon__hover );
+
+        border-width: 0;
+    }
+
+    .cf-icon {
+        margin-right: unit( 2px / @font-size, em );
+    }
+}
--- a/src/cf-typography/src/atoms/links.less
+++ b/src/cf-typography/src/atoms/links.less
@@ -1,0 +1,178 @@
+//
+// Links with icons
+// Required for jump link modifier
+//
+
+.a-link__icon,
+.a-link__jump {
+    border-bottom-width: 0;
+
+    .a-link_text {
+        border-bottom-width: 1px;
+        border-bottom-style: inherit;
+    }
+
+    &:before,
+    &:after {
+        .cf-icon();
+
+        margin-right: 1px;
+        margin-left: 1px;
+    }
+
+    &.a-link__download:after,
+    &.a-link__download&.a-link__before:before {
+        // Inserts the cf-icon-download glyph.
+        content: "\e406";
+    }
+
+    &.a-link__email:after,
+    &.a-link__email&.a-link__before:before {
+        // Inserts the cf-icon-email glyph.
+        content: "\e302";
+    }
+
+    &.a-link__external-link:after,
+    &.a-link__external-link&.a-link__before:before {
+        // Inserts the cf-icon-external-link glyph.
+        content: "\e610";
+    }
+
+    &.a-link__fax:after,
+    &.a-link__fax&.a-link__before:before {
+        // Inserts the cf-icon-fax glyph.
+        content: "\e310";
+    }
+
+    &.a-link__left:after,
+    &.a-link__left&.a-link__before:before {
+        // Inserts the cf-icon-left glyph.
+        content: "\e000";
+    }
+
+    &.a-link__mail:after,
+    &.a-link__mail&.a-link__before:before {
+        // Inserts the cf-icon-mail glyph.
+        content: "\e304";
+    }
+
+    &.a-link__pdf:after,
+    &.a-link__pdf&.a-link__before:before {
+        // Inserts the cf-icon-pdf glyph.
+        content: "\e402";
+    }
+
+    &.a-link__pdf:after,
+    &.a-link__pdf&.a-link__before:before {
+        // Inserts the cf-icon-pdf glyph.
+        content: "\e402";
+    }
+
+    &.a-link__phone:after,
+    &.a-link__phone&.a-link__before:before {
+        // Inserts the cf-icon-phone glyph.
+        content: "\e306";
+    }
+
+    &.a-link__right:after,
+    &.a-link__right&.a-link__before:before {
+        // Inserts the cf-icon-right glyph.
+        content: "\e002";
+    }
+
+    // New icon modifiers must be added ABOVE this line.
+
+    &.a-link__before {
+        &:after {
+            content: '';
+        }
+    }
+
+    &.a-link__no-wrap {
+        white-space: nowrap;
+    }
+}
+
+//
+// Jump link
+//
+
+.a-link__jump {
+    .webfont-medium();
+
+    &.a-link__large {
+        font-size: unit( @size-iv / @base-font-size-px, em );
+    }
+
+    .respond-to-max( @bp-xs-max, {
+        .a-link__block();
+
+        position: relative;
+
+        padding-right: 1.25em;
+
+        &:after,
+        &:before {
+            position: absolute;
+            margin-bottom: -0.6em;
+            bottom: 50%;
+        }
+
+        &:after {
+            right: 0;
+        }
+
+        &:before {
+            left: 0;
+        }
+
+        &.a-link__before {
+            padding-right: 0;
+            padding-left: 1.25em;
+        }
+
+        &.a-link__bg {
+            border: solid @jump-link_border;
+            border-width: 1px 0;
+            padding-left: 1em;
+            padding-right: 2.25em;
+
+            background: @jump-link_bg;
+
+            &:after {
+                right: 1em;
+            }
+
+            &:before {
+                left: 1em;
+            }
+
+            &.a-link__before {
+                padding-left: 2.25em;
+                padding-right: 1em;
+            }
+        }
+    } );
+}
+
+//
+// Block link
+//
+
+.a-link__block {
+    box-sizing: border-box;
+    display: block;
+
+    padding-top: unit( 10px / @base-font-size-px, em );
+    padding-bottom: unit( 10px / @base-font-size-px, em );
+    border-top-width: 1px;
+    border-bottom-width: 1px;
+    // 100% width is needed when block or jump link are applied to a <button>
+    width: 100%;
+
+    text-align: left;
+
+    .a-link_text {
+        border-bottom-width: 0;
+    }
+}
--- a/src/cf-typography/src/atoms/micro-copy.less
+++ b/src/cf-typography/src/atoms/micro-copy.less
@@ -1,0 +1,4 @@
+.a-micro-copy {
+    color: @micro-copy;
+    font-size: unit( @size-v / @base-font-size-px, em );
+}
--- a/src/cf-typography/src/cf-typography.less
+++ b/src/cf-typography/src/cf-typography.less
@@ -3,11 +3,6 @@
    Advanced Typography
    ========================================================================== */
 
-// Import external dependencies
-
-@import (less) '../../cf-core/src/cf-core.less';
-@import (less) '../../cf-icons/src/cf-icons.less';
-
 //
 // Theme variables
 //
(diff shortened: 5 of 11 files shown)


src/cf-typography/src/molecules/list.less
```diff
@@ -0,0 +1,105 @@
...
+        margin-top: unit( 24px / @base-font-size-px, em );
+    }
+}
+
+//
+// Horizontal list modifier
+//
+
+.m-list__horizontal {
+
```
Scotchester: trivial: remove blank line

src/cf-typography/src/molecules/list.less
```diff
@@ -0,0 +1,105 @@
...
+        margin-right: unit( 4px / @base-font-size-px, em );
+    }
+}
+
+//
+// Icon list modifier
+//
+
+.m-list__icons {
+
```
Scotchester: trivial: remove blank line

src/cf-typography/src/molecules/list.less
```diff
@@ -0,0 +1,105 @@
...
+        top: 0.1em;
+    }
+}
+
+//
+// Link list modifier
+//
+
+.m-list__links {
+
```
Scotchester: trivial: remove blank line

src/cf-typography/src/molecules/list.less
```diff
@@ -0,0 +1,105 @@
...
+    .a-list-item {
+
```
Scotchester: trivial: remove blank line

src/cf-typography/src/molecules/meta-header.less
```diff
@@ -0,0 +1,22 @@
+.m-meta-header {
+    padding-bottom: unit( 8px / @base-font-size-px, em );
+    // TODO: Move margin to o-post-preview and other parent containers
+    margin-bottom: unit( 10px / @base-font-size-px, em );
```
Scotchester: In the meantime, `margin` should come after `border`.

src/cf-typography/src/molecules/list.less
```diff
@@ -0,0 +1,105 @@
...
+        .respond-to-max( 599px, {
```
Scotchester: Use `@bp-xs-max`

src/cf-typography/src/molecules/list.less
```diff
@@ -0,0 +1,105 @@
...
+            margin-bottom: 0;
+
+            &:nth-child(n+2) .a-link {
+                border-top-width: 0;
+            }
+        } );
+    }
+
+    .a-link {
+        .respond-to-max( 599px, {
```
Scotchester: use `@bp-xs-max`

src/cf-typography/src/cf-typography.less
```diff
@@ -16,599 +11,56 @@
...
-            }
-            &:before {
-                left: unit(10px / @base-font-size-px, em);
-            }
-        }
-    });
-}
+// .m-slug-heading
+@slug-header_border__thin:  #323a45; // $color-gray-dark
+@slug-header_border__thick: #02bfe7; // $color-primary-alt
```
Scotchester: Trivial: double-check variable value alignment throughout.

src/cf-typography/usage.md
```diff
@@ -50,42 +50,39 @@ Default color variables are from 18F's
...
+@meta-header_border:        #323a45; // $color-gray-dark
 
 // .jump-link
-@jump-link__bg:                 #f1f1f1; // $color-gray-lightest
-@jump-link__bg-border:          #323a45; // $color-gray-dark
-
-// .list__branded
-@list__branded-bullet:          #02bfe7; // $color-primary-alt
+@jump-link_bg:              #f1f1f1; // $color-gray-lightest
+@jump-link_border:          #323a45; // $color-gray-dark
```
Scotchester: I'm not the biggest fan of doing a BEM-ish nomenclature for variable names, but if we do, shouldn't we keep it consistent with the class names things apply to? e.g., `pull-quote_body` should remain, since it applies to the element with that class (although the class is now prefixed with `m-`, and I certainly wouldn't want to add the prefixes to all variables...)

src/cf-typography/src/molecules/pull-quote.less
```diff
@@ -0,0 +1,26 @@
...
+        .heading-5();
+
+        color: @pull-quote_citation;
+    }
+
+    &__large .m-pull-quote_body {
+        .heading-2();
+
+        .respond-to-max( @bp-xs-max, {
+            .heading-3();
```
Scotchester: Why didn't we use `h#()` mixins here, so we didn't have to write the extra media queries?

src/cf-typography/usage.md
```diff
@@ -128,28 +125,28 @@ you should use the `aside` element.
...
-<aside class="pull-quote pull-quote__large">
-    <div class="pull-quote_body">
+<aside class="m-pull-quote m-pull-quote__large">
+    <p class="m-pull-quote_body">
         Lorem ipsum dolor sit amet, consectetur adipisicing elit.
         Cum corrupti tempora nam nihil qui mollitia consectetur
         corporis nemo culpa dolorum!
-    </div>
+    </p>
     <footer>
```
Scotchester: I'm not sure why we included this `footer` in the first place. Maybe kill it?

src/cf-typography/usage.md
```diff
@@ -50,42 +50,39 @@ Default color variables are from 18F's
...
-@jump-link__bg:                 #f1f1f1; // $color-gray-lightest
-@jump-link__bg-border:          #323a45; // $color-gray-dark
-
-// .list__branded
-@list__branded-bullet:          #02bfe7; // $color-primary-alt
+@jump-link_bg:              #f1f1f1; // $color-gray-lightest
+@jump-link_border:          #323a45; // $color-gray-dark
 ```
 
 ## Pull quote
```
Scotchester: Can we reorg this file to go from atoms to molecules?

src/cf-typography/usage.md
```diff
@@ -316,61 +265,89 @@ attached to `.meta-header_left` and `.category-slug` is attached to `.meta-heade
...
+        <span class="date">
+            Nov 4, 2013
+        </span>
+    </div>
+    <div class="m-meta-header_left">
+        <a href="#" class="m-meta-header_left category-slug">
+            <span class="cf-icon cf-icon-credit-card"></span>
+            Consumer finance
+        </a>
+    </div>
```
Scotchester: Why add wrappers?

Also, should be `a-date` and `m-category-slug` (here and below)

src/cf-typography/usage.md
```diff
@@ -316,61 +265,89 @@ attached to `.meta-header_left` and `.category-slug` is attached to `.meta-heade
...
 
-<div class="meta-header">
-    <a href="#" class="meta-header_right category-slug">
-        <span class="cf-icon cf-icon-credit-card"></span>
-    </a>
-    <span class="meta-header_left date">
-        Nov 4, 2013
-    </span>
+<div class="m-meta-header">
+    <div class="m-meta-header_left">
```
Scotchester: I feel like there might've been a reason why we always had put the `right` side first in the source order, but I'll test it out and see.

src/cf-typography/usage.md
```diff
@@ -566,12 +537,12 @@ It is primarily used within a max-width 599px media query (see `.jump-link` and
...
 #### Default example
 
-<a class="block-link" href="#">
+<a class="a-block-link" href="#">
     Default block link
 </a>
 
 ```
-<a class="block-link" href="#">
+<a class="a-block-link" href="#">
```
Scotchester: All of the link examples in this file are not using their full updated class names.

src/cf-typography/src/molecules/list.less
```diff
@@ -0,0 +1,105 @@
...
+    .a-list-item + .a-list-item {
+        margin-top: unit( 24px / @base-font-size-px, em );
+    }
+}
+
+//
+// Horizontal list modifier
+//
+
+.m-list__horizontal {
```
Scotchester: Why not continue including `__unstyled` in here? Not a fan of having to write the additional class in the markup.

src/cf-typography/src/atoms/headings.less
```diff
@@ -0,0 +1,31 @@
+.a-heading {
+
+    &__padded {
```
Scotchester: Why nest the modifiers in this file but not others? I'd go in the direction of un-nesting here, especially since `.a-heading` has no styles of its own.

src/cf-typography/src/molecules/list.less
```diff
@@ -0,0 +1,105 @@
+
```
Scotchester: Trivial: Remove blank line here, add comments at the beginning of each file.

src/cf-typography/src/atoms/links.less
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+
+//
+// Styled link
+//
+
+.a-link__styled {
+    .webfont-medium();
+}
```
jimmynotjim: I was hoping we could remove it and just use the medium font where we needed but wasn't sure. Consider it gone.

src/cf-typography/src/atoms/micro-copy.less
```diff
@@ -0,0 +1,8 @@
+.a-micro-copy {
+    color: @micro-copy;
+    font-size: unit( @size-v / @base-font-size-px, em );
+
+    &__large {
+        font-size: unit( @base-font-size-px / @base-font-size-px, em );
+    }
```
jimmynotjim: It's pre-existing and in use in a few places https://github.com/cfpb/cfgov-refresh/search?utf8=%E2%9C%93&q=micro-copy__large&type=Code

src/cf-typography/src/molecules/list.less
jimmynotjim: Removing duplicated code. I know gzipping takes care of some of this, but I've been trying to target duplication where possible so we can get to the cleanest output possible and then work our way back. These lists are used in so few templates, I think it's a good trade-off.

src/cf-typography/usage.md
jimmynotjim: Yep, that makes sense, went a bit overboard here.

src/cf-typography/usage.md
jimmynotjim: Argh, I reordered the sandbox but forgot this file.

src/cf-typography/src/molecules/pull-quote.less
jimmynotjim: because the `h#` mixins would include margins when these quotes but against other `h#` elems

src/cf-typography/src/atoms/headings.less
jimmynotjim: Good catch, I meant to nest them all, just so it's clear how they're all related, but I can unnest since there are no base styles.

src/cf-typography/usage.md
Scotchester: I think the alternate usage was mostly to demonstrate flopping the date and category, but honestly, it's probably not worth showing any alternate usage. It should be pretty self-evident what can be done with it.

src/cf-typography/src/molecules/pull-quote.less
Scotchester: That shouldn't ever happen in the context of a pull quote, though.

src/cf-typography/src/atoms/micro-copy.less
Scotchester: I searched and found those, too, but I still don't understand what the point is. It just ends up being the default body text of 16px Avenir. I vote to kill it.

src/cf-typography/usage.md
jimmynotjim: Because there's typically multiple categories, and it's only by chance this worked out beforehand. By separating the layout concerns from content elements, we can guarantee the layout will always work.

src/cf-typography/src/molecules/list.less
jimmynotjim: Ended up adding the horizontal list selector to the unstyled selector list. I think that'll keep things clean and bundled.

src/cf-typography/usage.md
Scotchester: Yeah, that's probably why. I agree; leave it.

src/cf-typography/src/molecules/list.less
```diff
@@ -1,9 +1,11 @@
-
 //
 // Unstyled list modifier
 //
```
Scotchester: This comment should maybe get slightly more detailed, with the addition of the other three selectors to the list.

src/cf-typography/usage.md
```diff
@@ -427,103 +331,94 @@ This is just another set of variations to show off what you could do.
...
-```
-<a class="a-styled-link" href="#">
-    Default styled link
-</a>
-```
-
 ### Jump link
 
-#### Default pattern
+### Jump link
```
Scotchester: Should remain h4; would not have it be the exact same wording as its parent heading.

Or... perhaps you meant to only have one of those.

src/cf-typography/usage.md
```diff
@@ -533,177 +428,171 @@ Jump links can also have icons before the text, like icon links.
...
 
 The icon list will set up a list with icons on the left, in place of bullets.
-Will need some more work to allow lines to wrap.
 
-<ul class="m-list m-list__unstyled m-list__icons">
-    <li class="m-list_item">
-        <span class="cf-icon cf-icon-email m-list_icon"></span>
-        List item 1
+<ul class="m-list m-list__icons">
+    <li class="a-list-item">
```
Scotchester: Wait, what?

src/cf-typography/src/atoms/headings.less
```diff
@@ -0,0 +1,28 @@
+.a-heading__padded {
```
Scotchester: This is no longer used anywhere. Off with its head!

src/cf-typography/src/molecules/list.less
```diff
@@ -0,0 +1,103 @@
...
+.m-list__unstyled,
+.m-list__horizontal,
+.m-list__icons,
+.m-list__links {
+    padding-left: 0;
+
+    list-style-type: none;
+
+    // This is needed for dd elements.
+    .a-list-item {
```
Scotchester: Missed this on my first review. What's the logic behind making this not an element of `.m-list` anymore?

src/cf-typography/src/molecules/list.less
```diff
@@ -0,0 +1,103 @@
...
+        .respond-to-max( @bp-xs-max, {
+            margin-bottom: 0;
+
+            &:nth-child(n+2) .a-link {
+                border-top-width: 0;
+            }
+        } );
+    }
+
+    .a-link {
```
Scotchester: Same question about it not being an element of the parent block anymore.

src/cf-typography/usage.md
```diff
@@ -153,207 +185,79 @@ you should use the `aside` element.
...
+### Heading with icon
 
-#### Default category slug
+The heading with icon is typically used for listing categories in a meta header.
+Since categories can be repetitive, we suggest placing a label with `.u-visually-hidden`
+prior to the headings to add more context for screen readers (see Meta Header below).
 
-<a href="#" class="m-category-slug">
-    <span class="m-category-slug_icon cf-icon cf-icon-dialogue"></span>
-    <span class="u-visually-hidden">Category:</span>
```
jimmynotjim: See the meta header section. When there are multiple, which a large portion of time there are, it reads `Category: Foo | Category: Bar` instead of `Categories: Foo | Bar`. I moved it to the meta header so that it'll be the latter and make more sense.

src/cf-typography/src/atoms/headings.less
jimmynotjim: really? Awesome.

src/cf-typography/src/molecules/list.less
jimmynotjim: Ah, that's what what. Because it leaves a lot more room to grow later. Yes it'll always be the child of a list, but by making it an atom we could globally target all item atoms that aren't just the default.

src/cf-typography/src/molecules/list.less
jimmynotjim: Because `a-link` is already an atom, we don't always need to chain the parent when working atomically. Nesting is perfectly fine in this instance because we're saying any time the atom link lives within the list link molecule.

src/cf-typography/usage.md
jimmynotjim: yeah, duplicated by accident

jimmynotjim:
@Scotchester Addressed everything with a comment or code change.

src/cf-typography/usage.md
Scotchester: I think that's a template implementation detail that should be left up to the user. We want to encourage the use of the hidden text when there's only one category present.

To play devil's advocate with myself, if this is no longer called a "category slug" but is called a "Heading with icon", perhaps the example should be changed to something else and we can avoid this issue entirely.

src/cf-typography/src/molecules/list.less
Scotchester: We could still do that with `.m-list_item`.

src/cf-typography/src/molecules/list.less
Scotchester: I guess I buy that in this case, but this is really throwing off our BEM game. I'm finding it hard to feel great about our overall reconciliation of atomic design and BEM.

src/cf-typography/src/molecules/pull-quote.less
Scotchester: > [W]hy add extra styles if we don't want them?

In order to simplify the experience of developing _this_ code. I am for the goal of slimming down output, but it has to be balanced with the goal of a good developer experience.

In this case, it's not making this code _so_ much more complex, so I will concede that it's probably worth it, in this case.

src/cf-typography/usage.md
jimmynotjim: Sorry I wasn't clear but yes even if there was only one, it should still be included in the meta header. The singular/plural implementation can be left to the template generating it. I updated the name to header with icon because I believe it follows our atomic design goals and makes it a reusable pattern outside of categories. What example do you suggest instead of an existing category?

src/cf-typography/src/molecules/list.less
Scotchester: I'm not following. Do you mean `.a-link.m-list_link`?

This is an awfully theoretical discussion, since `.a-link` has no styles and probably should never :)

jimmynotjim: Blah, I should have spent more time on formatting, I meant `class="a-list m-list_link` in the markup.

Scotchester: I think we're going to have to pair for me to understand what you're talking about here. 3–3:30?

src/cf-typography/src/atoms/date.less
```diff
@@ -0,0 +1,6 @@
+.a-date {
```
Scotchester: Should this go in `headings.less`? I don't want to encourage junk like what's happening in the right sidebar here: http://www.consumerfinance.gov/policy-compliance/rulemaking/final-rules/truth-lending-regulation-z-annual-threshold-adjustments-card-act-hoepa/

jimmynotjim: Hmm, it's not a heading, it's used in the meta area here => http://www.consumerfinance.gov/about-us/blog/lo-que-necesita-saber-para-recibir-dinero-del-acuerdo-de-compensacion-con-honda-finance-por-cobrarles-de-mas-las-minorias/

Maybe we just need to be clearer where it should and shouldn't be used?

jimmynotjim: EDIT: In the event details as well => http://www.consumerfinance.gov/about-us/events/fall-2016-consumer-advisory-board-meeting-washington-dc/

Scotchester: Yeah, I guess you're right. Ok, leave it here.

PR #413, opened 2016-10-14
Basic clean up of icons before atomic work

jimmynotjim:
Super small cleanup of icons before I start working with them. Started with the variables because I wanted to be sure I don't break anything in the process.
## Additions
- Added icons to link script
## Changes
- Updated unicode entries with variables
- Updated spacing to be consistent
## Testing
- Nothing really changed, but if you want to be sure the variables are correct, run `npm run cf-link` in this repo
- Open the sandbox and run `npm link cf-icons && npm run build && npm start`
- Open `http://localhost:3000/components/cf-icons/` and ensure the icons are all what they should be
## Review
- @Scotchester 
- @KimberlyMunoz 
## Notes
- This is only a pre-emptive change and while it may not be necessary once we move to svg, it'll be helpful for the time being.
## Checklist
- [x] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
- [x] New functions include new tests
- [x] Placeholder code is flagged
- [x] Visually tested in supported browsers and devices
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### DIFF
--- a/src/cf-icons/src/cf-icons.less
+++ b/src/cf-icons/src/cf-icons.less
@@ -19,8 +19,8 @@
 // IE7 Support
 //
 
-.cf-icon-ie7(@inner) when (@cf-icon-ie7-support) {
-    *zoom: ~"expression( this.runtimeStyle['zoom'] = '1', this.innerHTML = '@{inner}')";
+.cf-icon-ie7( @inner ) when ( @cf-icon-ie7-support ) {
+    *zoom: ~"expression(  this.runtimeStyle['zoom'] = '1', this.innerHTML = '@{inner}' )";
 }
 
 //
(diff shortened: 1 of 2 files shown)


src/cf-icons/src/cf-icons.less
```diff
@@ -44,24 +44,24 @@
...
-  //       https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth
-  -webkit-font-smoothing: antialiased;
+    font-family: 'CFPB Minicons';
+    display: inline-block;
+    font-style: normal;
+    font-weight: normal;
+    line-height: 1;
+    // TODO: Determine whether this can be safely removed and remove it.
+    //       Font smooth/smoothing is non-standard:
+    //       https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth
```
KimberlyMunoz: Can we test removing it on some IE browsers. I think so long as everything degrades gracefully (which I'm 95% sure it will), we should remove.

Scotchester: I'm probably also in favor of removing, but, as @KimberlyMunoz said, we should definitely do some testing to know exactly who/what it will affect. I don't want to just do it on the intuition that it'll probably be fine.

jimmynotjim: Why wouldn't it be fine? We are overwriting the default and only in Webkit. No other browser has ever supported the font-smoothing property. IE isn't even affected, only Safari and Chrome.

Scotchester: Ok, I've done some quick testing.

1. Chrome on Windows is _not_ affected. I think because it uses DirectWrite for all text rendering on Windows.
   - Safari hasn't been supported on Windows since v5.
2. There are a lot of Mac users that don't have Retina Displays, or regularly plug into a non-Retina display.
3. You might want to check your assumptions about what looks good. Open up the sandbox in Chrome, scroll to the YouTube icon, inspect it, and flick the `-webkit-font-smoothing` on and off. Try both on your Thunderbolt Display and on the Retina Display.

Unless it's causing an issue, my vote at this time is to leave it.

jimmynotjim: I had checked on both retina and non-retina (Thunderbolt Display and personal machines) but hadn't noticed the small type in the YouTube icon. Every other icon is crisper at the default setting, but it's probably not worth investigating more at this point. Instead I'm going to remove the `TODO` comment so no one else goes down this rabbit hole.

Scotchester: Yeah, the simple things like arrows look marginally better, but stuff with fine detail can fall apart, and we have lots of those in our financial products section, especially.

PR #415, opened 2016-11-01
Look out! It's Atomic Expandables!

mistergone:
This PR updates `cf-expandables` to use the AtomicComponent package, which removes its dependency on jQuery. It also updates the LESS to an atomic standard.

## Important Notes (Read First!)
- Some of this PR is still in-progress, so you might find some things lacking. It works, but please speak up if I missed a detail or some documentation!
- For IE8, this code does not support JS. Instead, all expandables are simply open, and the "cues" (Open/Close, and icons) are hidden.
- For IE9, CSS transitions are not supported, so expandables open and close, but there is no transition animation

## Additions
- Adds the `AtomicComponent` package and applies it to expandables

## Removals
- Removes jQuery code

## Review
- @Scotchester @jimmynotjim and anyone else interested

### To get this running
- `git clone` (or whatever) this branch
- Assuming you're testing with `capital-framework-sandbox`, you're going to want to use __[this branch right here](https://github.com/cfpb/capital-framework-sandbox/tree/atomicjs-exp)__

## Screenshots
[omitted - no visual changes]

## Notes
- See PR description

## Todos
- Analytics handlers inside the expand and collapse function would be helpful!

## Checklist


src/cf-expandables/src/ExpandableToggle.js
```diff
@@ -0,0 +1,92 @@
...
+
+var contains = require( 'atomic-component/src/utilities/dom-class-list' ).contains;
+var addClass = require( 'atomic-component/src/utilities/dom-class-list' ).addClass;
+var closest = require( 'atomic-component/src/utilities/dom-closest' ).closest;
+var bind = require( 'atomic-component/src/utilities/function-bind' ).bind;
+var removeClass = require( 'atomic-component/src/utilities/dom-class-list' ).removeClass;
+var ExpandableTransition = require( 'atomic-component/src/utilities/transition/ExpandableTransition' );
+var Events = require( 'atomic-component/src/mixins/Events.js' );
+
+var ToggleExpandable = {
```
sebworks: This doesn't have to be a modifier. It can just be a part of `Expandable.js`.

src/cf-expandables/src/ExpandableToggle.js
```diff
@@ -0,0 +1,92 @@
...
+
+  transition: null,
```
sebworks: Should be set to `UNDEFINED`. Typically only use null when setting variables that will reference dom elements.

src/cf-expandables/src/cf-expandables.js
```diff
@@ -1,117 +1,8 @@
...
- */
+'use strict';
 
-global.jQuery = require('jquery');
-require('jquery.easing');
+var Expandable = require( './Expandable' );
 
-(function( $ ) {
+// polyfill for ie9 compatibility
+require( 'classlist-polyfill' );
```
sebworks: The classlist utility should work without the polyfill.

src/cf-expandables/src/ExpandableToggle.js
mistergone: I moved all the code from `ExpandableToggle.js` into `Expandable.js` and it all works fine. Thanks for the suggestion of simplification!

src/cf-expandables/src/ExpandableToggle.js
mistergone: Good call. I wasn't really thinking about `null` versus `undefined`. Thanks!

src/cf-expandables/src/cf-expandables.js
mistergone: Do you mean `AtomicComponent/src/utilities/dom-class-list/`? Maybe the way I'm `require`-ing it is preventing it from working properly? I was having problems in IE9 that fixed themselves with the polyfill, but I'm all ears on an easier way to fix them. 😄

sebworks: ok, I'll check it on IE9

src/cf-expandables/src/cf-expandables.less
```diff
@@ -16,186 +16,186 @@
...
 //
 // Recommended expandable pattern
 //
 
-.expandable {
-    margin-bottom: unit(30px / @base-font-size-px, em);
-}
+.o-expandable {
+    position: relative;
+    margin-bottom: unit( 30px / @base-font-size-px, em );
```
jimmynotjim: Ideally the spacing between expandables should be controlled by it's parent template.

src/cf-expandables/src/cf-expandables.less
```diff
@@ -16,186 +16,186 @@
...
+            outline: 1px dotted @expandable-focus;
+            outline-offset: 1px;
+        }
 
-.expandable_content {
-    // A clearfix prevents twitchy animations from occurring when margins
-    // collapse and extend past the bounds of the expandable.
-    .u-clearfix();
-}
+        &__expanded .o-expandable_cue-open {
```
jimmynotjim: Could we use a psedo elem that changes it's content for the cues instead of elements that are hidden/shown?

jimmynotjim:
I think you need to push your changes to the sandbox, the expandables still have the old class names in the branch referenced above.

mistergone:
@Scotchester @jimmynotjim and anyone else:

I've updated the PR's text because my changes are now live in `atomic-component`, so all you should need is this branch of `capital-framework` and the specified branch of `capital-framework-sandbox`.

Please try it out and let me know!

jimmynotjim:
Yay, got it working locally. I'd suggest speeding up the animations, they feel really sluggish. George and I found between 0.2 and 0.3 seconds to be about right for animations like this when working on the multiselect drop down.

mistergone:
@jimmynotjim - Easy change in the CSS (which also allows anyone to change it in their own implementations!). I'll pop it down to `0.25` for purposes of default setting.

jimmynotjim:
@mistergone It's not a variable, how would you change it? Thinking about it now, maybe it should be a variable so it's easily over-written in a theming file.

Scotchester:
I also like the idea of making it a variable.

mistergone:
@jimmynotjim - Oh, I'm just thinking in the project CSS, it could be changed to whatever. (Previously, when I was doing it in JavaScript, it would've been much more involved to change it.)

But a variable makes lots of sense.

jimmynotjim:
Added the variable while I had a second between tasks.

src/cf-expandables/src/cf-expandables.less
jimmynotjim: Realized this is the existing implementation, let's address this at a later date.

src/cf-expandables/src/cf-expandables.js
jimmynotjim: @sebworks did you get a chance to check this?

src/cf-expandables/src/cf-expandables.less
jimmynotjim: Realized while pairing expandables aren't always in groups. Let's refactor this later.

jimmynotjim:
Tested again locally and works great. Thanks again for seeing this through @mistergone.

For anyone testing: If you run into an error where the expandables are all open and not interactive followed by this in the console
```
Uncaught Error: Cannot find module "classlist-polyfill"
```
delete `capitalframework/src/cf-expandables/npm_modules` and re-run `npm run cf-link` in the both repos. It should resolve any issues.

PR #416, opened 2016-11-09
Big atomic update to cf-icons

jimmynotjim:
Default usage of icons required an empty elem because the font family was set on the elem level. This change restricts the font-family to the pseudo-elem allowing icons to be used with standard text. This change will eliminate the duplicated icon classes in cf-buttons and cf-typography. Icon only elems can still be used by adding `cf-icon__before` the existing class list.

## Changes

- Updated the `cf-icon` class to be a mixin only, so that it's not duplicated when `cf-icon-prefix` is set as `cf-icon`.
- Updated the pseudo elem to only include an icon when a modifier class is used, this is important to mix icons into elems with normal text like buttons, links, etc.
- Removed IE7 support because the hack stopped working.

## Testing

- check out this branch and run `npm run cf-link`
- check out Sandbox branch `example/cf-icons` and run `npm run cf-link`
- run `npm start`
- check `http://localhost:3000/components/cf-icons/`

## Review

- @Scotchester 
- @KimberlyMunoz 

## Screenshots

Everything's the same

## Notes

- Adding the `__before` modifier is a small but necessary hurdle to making this possible. The benefit of removing 20kb of duplicated code outweighs that extra work.

## Todos

- Remove the duplicated icon code through the rest of the project.

## Checklist

* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG) # #

### DIFF
--- a/src/cf-icons/src/cf-icons.less
+++ b/src/cf-icons/src/cf-icons.less
@@ -8,20 +8,17 @@
 @import '../../cf-core/src/cf-core.less';
 
 //
-// Less variables
+// Theme variables
 //
 
 @cf-icon-prefix:                cf-icon;
 @cf-icon-path:                  'fonts';
-@cf-icon-ie7-support:           true;
-@cf-icon-border-color:          #5b616b; //$color-gray in 18F's web design standards
-//
-// IE7 Support
-//
 
-.cf-icon-ie7( @inner ) when ( @cf-icon-ie7-support ) {
-    *zoom: ~"expression(  this.runtimeStyle['zoom'] = '1', this.innerHTML = '@{inner}' )";
-}
+// Color variables
+// $color- variables are from 18F's U.S. Web Design Standards
+// https://github.com/18F/web-design-standards/blob/18f-pages-staging/src/stylesheets/core/_variables.scss
+
+@cf-icon-border-color:          #5b616b; //$color-gray
 
 //
 // The basics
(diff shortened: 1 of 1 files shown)


KimberlyMunoz:
I'm not a big fan of the before/after classes as opposed to the empty elements. The empty span elements can be controlled with an aria-hidden to avoid reading nonsense to a screenreader, but using pseudo-elements is a little riskier. Hopefully when y'all switch to SVGs this won't be a big deal, but until then, there's some accessibility concerns I'd test with Jennifer Horan.

jimmynotjim:
@KimberlyMunoz If we were to continue using the empty elems with the `__before` class, but removed the already in use psuedo specific elems in buttons and links in favor of this approach, do you see any accessibility issues?

KimberlyMunoz:
Someone would have to test, but my guess is that it probably won't be worse than what we have now. The icon fonts just do funny things with what they read out to screenreaders.

jimmynotjim:
Yeah, my hope is once we move to svg, we'd have the option of using the image with either the content or background of the pseudo-elem and solve any accidental screen readings. I'm glad you pointed out the aria rule, it doesn't look like we're currently using it on the empty icon elems in production. It's definitely something we need to fix.

I've tested with VoiceOver and that currently skips the icons. I'll try and set something up with Jennifer to test with Windows.

KimberlyMunoz:
Thanks! I know this is something that still has a ways to go, so I appreciate the progress being made here!

jimmynotjim:
@Scotchester Fixed the issue in your last comment by creating a reusable mixin that contains all of the chained pseudo elems

src/cf-icons/src/cf-icons.less
```diff
@@ -8,12 +8,17 @@
...
 
 @cf-icon-prefix:                cf-icon;
 @cf-icon-path:                  'fonts';
-@cf-icon-border-color:          #5b616b; //$color-gray in 18F's web design standards
+
+// Color variables
+// $color- variables are from 18F's U.S. Web Design Standards
+// https://github.com/18F/web-design-standards/blob/18f-pages-staging/src/stylesheets/core/_variables.scss
+
+@cf-icon-border-color:          #5b616b;
```
jimmynotjim: Whoops, got overly zealous with deleting

PR #421, opened 2016-12-01
Adding the forgotten document changes and Changelog entry

jimmynotjim:
@Scotchester You warned me, and I should know better by now, but I still forgot the CHANGELOG and documentation. This fixes that.

### DIFF
--- a/src/cf-icons/usage.md
+++ b/src/cf-icons/usage.md
@@ -30,7 +30,8 @@ This component can be used by itself, but is designed to work with Capital Frame
 
 ## Variables
 
-Theme variables for setting the color and sizes throughout the project. Overwrite them in your own project by duplicating the variable `@key: value`.
+Theme variables for setting the color and sizes throughout the project.
+Overwrite them in your own project by duplicating the variable `@key: value`.
 
 ### Settings
 
@@ -44,7 +45,8 @@ Theme variables for setting the color and sizes throughout the project. Overwrit
 
 The cf-icon-prefix class applies all shared icon styles including the font family.
 By default, this class will be named `cf-icon` but it can be changed in the settings.
-All icons must use three classes, one for this base class, one for the chosen pseudo-element, and one to set the font character. For example:
+All icons must use three classes, one for the base class, one to select the
+desired icon and one for the chosen pseudo-element. For example:
 
 ```
 <span class="cf-icon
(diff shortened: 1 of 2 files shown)


src/cf-icons/usage.md
```diff
@@ -37,33 +37,22 @@ Theme variables for setting the color and sizes throughout the project. Overwrit
...
-@cf-icon-ie7-support: false;
-```
-
-
 ## The basics
 
 The cf-icon-prefix class applies all shared icon styles including the font family.
 By default, this class will be named `cf-icon` but it can be changed in the settings.
-All icons must use two classes, one for this base class and another to set the font character. For example:
+All icons must use three classes, one for this base class, one for the chosen pseudo-element, and one to set the font character. For example:
```
Scotchester: Minor: `set the fond character` ➡️ `select the desired icon`. Setting the font character is an implementation detail they can discover by looking at the source, if they're really curious.

src/cf-icons/usage.md
```diff
@@ -37,33 +37,22 @@ Theme variables for setting the color and sizes throughout the project. Overwrit
...
 
 ```
-<span class='cf-icon cf-icon-left'></span>
+<span class="cf-icon
+             cf-icon-money
+             cf-icon__before"></span>
 ```
 
-Please use the span element instead of the i element.
+When using an empty element, please use the span element instead of the i element.
```
Scotchester: Should we say here that using empty elements is now discouraged when avoidable?

jimmynotjim: Agreed and updated.

src/cf-icons/usage.md
jimmynotjim: Addressed all of these.

src/cf-icons/usage.md
```diff
@@ -44,16 +45,18 @@ Theme variables for setting the color and sizes throughout the project. Overwrit
 
 The cf-icon-prefix class applies all shared icon styles including the font family.
 By default, this class will be named `cf-icon` but it can be changed in the settings.
-All icons must use three classes, one for this base class, one for the chosen pseudo-element, and one to set the font character. For example:
+All icons must use three classes, one for the base class, one to select the
+desired icon and one for the chosen pseudo-element. For example:
```
Scotchester: Please insert comma between "icon" and "and".

src/cf-icons/usage.md
```diff
@@ -44,16 +45,18 @@ Theme variables for setting the color and sizes throughout the project. Overwrit
...
 ```
 <span class="cf-icon
              cf-icon-money
              cf-icon__before"></span>
 ```
 
-When using an empty element, please use the span element instead of the i element.
-This avoids font family cascading conflicts when using an italic webfont
+It's preferred to combine the icon classes with an existing element, but if it's
+necessary to use an empty element, please use the span element instead of the i
```
Scotchester: It'd be nice to wrap the "span" and "i" in backticks for code formatting.

src/cf-icons/usage.md
```diff
@@ -44,16 +45,18 @@ Theme variables for setting the color and sizes throughout the project. Overwrit
...
+element. This avoids font family cascading conflicts when using an italic webfont
 on i elements and then another font for the icons.
```
Scotchester: Also here.

jimmynotjim:
@Scotchester addressed your comments and ready for sign-off.

PR #424, opened 2016-12-30
WIP: Adds custom-input atomic components

anselmbradford:
Updates custom radio and checkbox inputs and input groups to support multi-line labels and small helper text in the label.

## Changes

- Adds flexbox with table-layout fallback to `.m-form-field`.
- Increases radio/checkbox size to 20px to match latest spec.
- Uses box-shadow in place of outline so size of button doesn't increase when clicked.


jimmynotjim:
@anselmbradford do you think it's better to duplicate what we're about to [release in ver4](https://github.com/cfpb/capital-framework/blob/prerelease-ver-4/src/cf-forms/src/molecules/form-fields.less) so that there's no need to make any changes later?

src/cf-forms/src/molecules/form-fields.less
```diff
@@ -11,15 +12,19 @@
...
 
             &:before {
-                display: inline-block;
+                display: table-cell;
+                align-self: flex-start;
 
                 border: 1px solid @input-border;
-                height: unit( 18px / @base-font-size-px, em );
-                width: unit( 18px / @base-font-size-px, em );
+                min-width: unit( 20px / @base-font-size-px, em);
```
jimmynotjim: Width will over-write this.

src/cf-forms/src/molecules/form-fields.less
```diff
@@ -123,7 +131,7 @@
         .a-radio {
             &:focus + .a-label,
             &:checked + .a-label {
-                outline: 1px solid @input-border__focused;
+                box-shadow: 0px 0px 0px 1px @input-border__focused inset;
```
jimmynotjim: Curious why the change to box-shadow?

src/cf-forms/src/molecules/form-fields.less
```diff
@@ -11,15 +12,19 @@
     &__checkbox,
     &__radio {
         .a-label {
+            display: table-row;
```
jimmynotjim: Does the parent need to be `display: table` for this to work properly?

src/cf-forms/src/cf-forms.less
```diff
@@ -10,43 +10,45 @@
...
 
+@input-disabled:                #75787b; // @gray
 
 // Sizing variables
 
 @select-height:                  30px;
 
 // Import external dependencies
 
+@import (less) '../../cf-core/src/cf-base.less';
```
jimmynotjim: Discussed in chat, let's figure out why `cf-core` isn't pulling this in correctly.

src/cf-forms/src/molecules/form-fields.less
anselmbradford: No it's not necessary unless defining a specific structure that requires it. The table display classes will create what they need.

src/cf-forms/src/molecules/form-fields.less
anselmbradford: Outline adds to the outside of the box, increasing its size by 2px in width and height. box-shadow puts the line inside the box.

src/cf-forms/src/molecules/form-fields.less
anselmbradford: Hmm, for some reason I needed this to prevent flexbox from stretching the size of the checkbox/radio icon, but seems to work now without it.

src/cf-forms/src/molecules/form-fields.less
anselmbradford: Hrm, not working quite as expected, but I think for IE9 I'm not going to worry about it.

screen shot 2017-01-05 at 6 49 39 pm

Also, the color is off in IE9. Same thing for the footer color.

src/cf-forms/src/molecules/form-fields.less
Scotchester: That (adding to the outside of the box) is the intended design.

src/cf-forms/src/molecules/form-fields.less
jimmynotjim: Could you leave a note just so no one thinks it can be removed in the future.

src/cf-forms/src/molecules/form-fields.less
jimmynotjim: Yeah, that's what I thought. Let's keep this to stick with the existing design.

anselmbradford: Ah, I had mixed messages on this. Do we definitely have design docs that the selected size should be larger? Meaning a line of vertically stacked radio buttons will not have an aligned left/right edge when one is selected.

jimmynotjim:
@anselmbradford what ever happened with this? Are you still looking to make changes to the current large target behavior or can we close this PR?

anselmbradford:
@nataliafitzgerald Yeah, unfortunately as you can see I haven't had a chance to circle around back to it. Currently the big inputs don't support what is in use in the Mosaic project.

anselmbradford:
Closed in favor of https://github.com/cfpb/capital-framework/pull/631

PR #434, opened 2017-01-11
Remove box-sizing polyfill

anselmbradford:
The box-sizing polyfill is only for IE6/7 and hasn't been updated in three years. I think it's time to remove it and free ourselves of it as a dependent setting.

## Removals

- Remove references to the box-sizing polyfill.

## Testing
- Run `npm run build` and check that IE8+ works in the CF sandbox.

## Notes

- Updates to cf-form are purely to allow the build step to pass.

### DIFF
--- a/src/cf-grid/src/cf-grid.less
+++ b/src/cf-grid/src/cf-grid.less
@@ -12,7 +12,6 @@
 // Less variables
 //
 
-@grid_box-sizing-polyfill-path: '../../box-sizing-polyfill';
 @grid_wrapper-width:            1200px;
 @grid_gutter-width:             30px;
 @grid_total-columns:            12;
@@ -60,14 +59,6 @@
   .lt-ie8 & {
     // Hack inline-block into submission
     display: inline;
-    // Mixing negative margins and the box-sizing hack can be dangerous.
-    // This is a temporary fix for: https://github.com/cfpb/cf-grid/issues/17
-    margin-right: 0;
-
-    // Clearfix hack for IE 6/7 only
-    zoom: 1;
-    // Box-sizing polyfill
-    behavior: url('@{grid_box-sizing-polyfill-path}/boxsizing.htc');
   }
 
   // Modifying standard width and padding for prefixed/suffixed columns, if necessary:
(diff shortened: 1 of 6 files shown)


src/cf-forms/src/cf-forms.less
```diff
@@ -10,36 +10,37 @@
 
 // Input borders
 @input-border:                  #5b616b;
-@input-border__hover:           @pacific;
-@input-border__focused:         @pacific;
-@input-border__active:          @pacific;
-@input-border__error:           @red;
-@input-border__warning:         @gold;
-@input-border__success:         @green;
+@input-border__hover:           #0072ce; // @pacific;
```
jimmynotjim: Remove these variable changes so that it doesn't clash with your other PR.

src/cf-grid/src/cf-grid.less
```diff
@@ -60,14 +59,6 @@
   .lt-ie8 & {
     // Hack inline-block into submission
     display: inline;
-    // Mixing negative margins and the box-sizing hack can be dangerous.
-    // This is a temporary fix for: https://github.com/cfpb/cf-grid/issues/17
-    margin-right: 0;
-
-    // Clearfix hack for IE 6/7 only
-    zoom: 1;
```
jimmynotjim: I think we still need the zoom hack right?

src/cf-forms/src/cf-forms.less
anselmbradford: It won't build otherwise. I can add them as a separate (third PR), since any PR off of v4 branch will run into this.

src/cf-grid/src/cf-grid.less
anselmbradford: I was going off the comment saying it was IE 6/7 only, do you know what we need it for?

jimmynotjim: No, just being careful because the comment didn't mention the polyfill like the other changes. Maybe @Scotchester does?

src/cf-forms/src/cf-forms.less
jimmynotjim: Leave it and figure out the collusion when you merge the second one.

PR #437, opened 2017-01-12
Updated the changelog for the v4 release

jimmynotjim:
Preparing for the v4 release. Condensed the changes/additions but kept the deprecated removals for user reference.


Scotchester:
The only comment I have is keep this diff handy. I think people doing migrations to v4 will want to reference some of the details you removed from the Changes section.

PR #450, opened 2017-02-23
Update list spacing for latest from DM team

jimmynotjim:
Moved @scotchester and @nataliafitzgerald's work from https://github.com/cfpb/design-manual/issues/357 to this project.

## Additions

- Nested list spacing and style

## Removals

- Old list spacing

## Testing

- Checkout `update-lists` [branch in the sandbox](https://github.com/cfpb/capital-framework-sandbox/tree/update-lists).
- Run `npm run cf-link` here and in the sandbox.
- Run `npm run build && npm start` in the sandbox.
- Navigate to `http://localhost:3000/components/cf-core/cf-base.html`.

## Review

- @Scotchester 
- @nataliafitzgerald

## Screenshots

screen shot 2017-02-23 at 11 28 27 am

## Checklist

### DIFF
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -377,6 +377,11 @@ p + ol {
     margin-top: unit( -5px / @base-font-size-px, em );
 }
 
+ul ul,
+ol ul {
+    margin-top: unit( 8px / @base-font-size-px, em );
+}
+
 li {
     margin-bottom: unit( 8px / @base-font-size-px, em );
 
@@ -450,20 +455,16 @@ nav a {
 // Lists
 //
 
+ul, ol {
+    padding-left: unit( 18px / @base-font-size-px, em );
+}
+
 ul {
-    // 32px provides exactly 15px between left margin
-    // and bullet character in most browsers
-    padding-left: unit( 32px / @base-font-size-px, em );
     list-style: square;
 }
 
-// Use the same overall number as ULs, but pull the number 1px to the left
-ol {
-    padding-left: unit( 31px / @base-font-size-px, em );
-
-    > li {
-        padding-left: unit( 1px / @base-font-size-px, em );
-    }
+ul ul {
+    list-style-type: circle;
 }
 
 //
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -370,6 +370,10 @@ table,
 blockquote {
     margin-top: 0;
     margin-bottom: unit( 15px / @base-font-size-px, em );
+
+    &:last-child {
+        margin-bottom: 0;
+    }
 }
 
 p + ul,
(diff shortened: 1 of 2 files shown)


src/cf-core/src/cf-base.less
```diff
@@ -383,6 +383,11 @@ p + ol {
     margin-top: unit( -5px / @base-font-size-px, em );
 }
 
+ul ul,
+ol ul {
```
Scotchester: YOu forgot to add `ol ol` here. And, I suppose, `ul ol`.

nataliafitzgerald:
@jimmynotjim @Scotchester 
When we have secondary bullets what is the space between the secondary bullet and the primary bullet that follows? In the screenshot above it appears that the space between the secondary bullet and the primary bullet that follows matches the space between paragraphs. Instead this space should match the space between primary bullets. This looks like it was correct in the [Codepen](http://codepen.io/Scotchester/pen/rrokQE?editors=1100) that @Scotchester created.

Scotchester:
@nataliafitzgerald Hmm, it does look like that, @nataliafitzgerald. The CSS in the pen, which inherits from the live CSS on consumerfinance.gov, does do this correctly.

jimmynotjim:
@nataliafitzgerald @Scotchester inspecting in the sandbox, it's 15px bottom `margin` it's inheriting in [this global list rule on L:371](https://github.com/cfpb/capital-framework/pull/450/files#diff-7049f34d37b91f0fc0ebe1c025293fdeR371)

src/cf-core/src/cf-base.less
jimmynotjim: Whoops, thanks for catching that.

Scotchester:
@jimmynotjim The question is why is it not getting the `:last-child` 'margin-bottom: 0;` rule? I'm guessing that's only in cfgov-refresh at this point.

jimmynotjim:
@Scotchester I'm not seeing a `:last-child` rule anywhere in here other than the one for nav list items. I'm also not seeing it in the `unprosessed` files in cfgov-refresh. Do you think we should apply it to the [list here](https://github.com/cfpb/capital-framework/blob/canary/src/cf-core/src/cf-base.less#L364-L373)

jimmynotjim:
Wait, I found it, a bit buried. https://github.com/cfpb/cfgov-refresh/blob/83c1281bd92868437c5a8b611d686f020d53f2dc/cfgov/unprocessed/css/enhancements/typography.less#L386-L390 It seems as though this should apply to the entire list I referenced above.

Scotchester:
> It seems as though this should apply to the entire list I referenced above.

Agreed.

jimmynotjim:
Fixed:

screen shot 2017-02-23 at 1 39 45 pm

Scotchester:
Looks good, @jimmynotjim. I'd mention that update as a separate item in the change log, and then this is good to go.

Scotchester:
Fixed `CHANGELOG.md` conflict. merging when Travis finishes.

PR #451, opened 2017-02-28
CF Release

CFPBot:
### Changed
- **all components:** Standardized all the usage docs for v4
- **cf-core:** Updated the base list styles to match latest DM Spec
- **cf-core:** Updated the body copy elements to remove margin from the `last-child`
- **all components:** Corrected the variables correct names, locations, and organization

### Removed
- **all components:** Deprecated code
  - Deprecated webfont mixins
  - Extra select element

## Review

- @cfpb/front-end-team-admin

I am a [bot](https://github.com/cfpb/cfpbot). If I did something wrong, blame a human.

kitten gif


Scotchester:
One problem we have with reviewing these PRs is that there is no way to tell from just this PR if the version bumps should be major/minor/patch.

src/cf-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cf-core",
-  "version": "4.0.1",
+  "version": "4.2.0",
```
Scotchester: We also still have this issue where it bumps two increments in a single release.

jimmynotjim: Ah heck. I really wonder if we need to do condensed Changelog entries right before opening a bot PR when there are multiple changes like this.

contolini:
Closing to test a bot fix.

PR #289, opened 2016-02-24
MVP of including component usage.md files in the documentation

ascott1:
This is an MVP of getting the `usage.md` files moved into the documentation.

What this does:

- Adds a gulp task that takes the component usage.md and converts it to html so that it can be imported into a Jekyll template
- Does the importing and sets up navigation for those items
- Adds the cf `js` file so that the expandables demo works
- Adds dummy classes for the `cf-grid` demo
- Shortens the text on the landing page (scope creep, but ever since @cfarm mentioned it, it's really been bugging me)

## Demo

https://ascott1.github.io/capital-framework/
https://ascott1.github.io/capital-framework/components/

## Review

Any two of the following: @contolini @Scotchester @KimberlyMunoz @jimmynotjim @cfarm

PR #454, opened 2017-03-01
Bugfix for GitHub flavored Markdown

jimmynotjim:
GitHub style heading links weren't working. Turned out we were converting the files to html outside of kramdown. Updated Gulp to only copy the Markdown files and the Kramdown config to use the GitHub style settings.

## Removals

- Removed the Gulp usage task and related packages

## Changes

- Added usage docs to the Gulp copy task
- Updated Jekyll config to use GitHub settings in Kramdown

## Testing

- If you've never run the gh-pages branch, run `npm install`
- Navigate to the components and check that they display properly, including GitHub style code blocks

## Review

- @ascott1 
- @Scotchester 

## Checklist


gulp/tasks/build.js
```diff
@@ -7,7 +7,6 @@ gulp.task( 'build',
     'styles',
     'scripts',
     'images',
-    'copy',
-    'usage'
+    'copy'
```
Scotchester: Trivial: You could leave the comma on this line to reduce the diff to one line, here.

gulp/tasks/copy.js
```diff
@@ -36,10 +38,30 @@ gulp.task( 'copy:vendorjs', function() {
...
+    .pipe( $.changed( config.usage.dest ) )
+    .pipe( $.tap( function ( file, t ) {
+      fileName = path.dirname( file.path )
+        .split( path.sep )
+        .pop();
+    } ) )
+    .pipe( $.rename( function ( path ) {
+      path.basename = fileName;
+      path.extname = '.md'
+    } ) )
```
Scotchester: What's all this about? Isn't this PR proposing just copying straight `*.md` to `*.md` files?

gulp/tasks/copy.js
```diff
@@ -36,10 +38,30 @@ gulp.task( 'copy:vendorjs', function() {
...
+    } ) );
+} );
+
 gulp.task( 'copy',
   [
     'copy:files',
     'copy:icons',
-    'copy:vendorjs'
+    'copy:vendorjs',
+    'copy:usage'
```
Scotchester: Trivial: Could add a comma here to prevent the need to modify this line in the future when adding something else to the list.

Scotchester:
Responding to myself, the original PR, #289, said:

> Adds a gulp task that takes the component usage.md and converts it to html so that it can be imported into a Jekyll template

I'll test it myself, though I'm sure you tested this and that it works, @jimmynotjim. Perhaps a Jekyll update made it possible to import Markdown files?

gulp/tasks/copy.js
jimmynotjim: Yes, but every file name is `usage.md`. You still need to rename it to match the parent package.

jimmynotjim:
@Scotchester not sure, this answer is from '13 => http://stackoverflow.com/questions/14668403/how-to-render-a-jekyll-markdown-page-on-sites-index

I have a feeling it might have been due to the need to set the Kramdown settings. Before I figured that out the output wasn't the same as GitHubs.

gulp/tasks/build.js
jimmynotjim: Our [linting settings say](https://github.com/cfpb/development/blob/master/.eslintrc#L560) otherwise

gulp/tasks/copy.js
Scotchester: Ahhh, gotcha.

gulp/tasks/build.js
Scotchester: > This rule is needed for IE8- support.

Booooo. We should have an exception for code that doesn't run in-browser, but I don't know how we'd handle that on a practical level.

jimmynotjim: I disagree. I don't want to have to know different rules for different environments.

Scotchester: I don't want unnecessary diffs 😝

Scotchester: Don't we not serve JS to IE8 at this point, anyway? If so, we could change this rule to the correct value of `always`.

jimmynotjim: @Scotchester Out of scope but lets bring it up in `cfpb/development`

jimmynotjim:
@Scotchester Good catch, I should have scrolled more of the pages. I have a fix ready, cf-core was the only one that had this issue. Only problem is that Travis still hasn't kicked off the release, so I can't edit the CHANGELOG without creating a conflict. We're stuck waiting till that's all flushed out.

ascott1:
> I have a feeling it might have been due to the need to set the Kramdown settings. Before I figured that out the output wasn't the same as GitHubs.

Yes, I believe this is correct.

PR #465, opened 2017-03-31
Update Travis script to print which components are published to npm

contolini:
Travis prints all sorts of messages during its npm prepublish script but is silent during its publishing step.

See: https://[mm]/cfpb/pl/tyawf6wfhtbedguopa7w39nawy

## Additions

- Print the name of every component that's about to be published.

## Review

- @Scotchester 
- @jimmynotjim


jimmynotjim:
Looks good 👍

PR #467, opened 2017-04-04
Call for Discussion - Fix a bug when animating icons within a button

jimmynotjim:
Currently, if you add the `cf-icon__spin` class to an icon within a button, it spins everything, the padding, border, and margin due to those properties being attached to the icon pseudo element. 

screen shot 2017-04-03 at 1 11 59 pm

In order to spin the icon without spinning the other properties, we have to separate the two. This change won't affect icon usage anywhere else, only this specific case (meaning we would only need the empty spans here).

## Changes

- Updated buttons with icons styles and markup pattern

## Testing

1. Clone the this branch to ~/Downloads/capital-framework.
2. cd ~/Downloads/capital-framework and npm install && npm run cf-link
3. Clone the gh-pages branch to ~/Downloads/cf-site.
4. cd ~/Downloads/cf-site and npm install && npm run cf-link && npm run build && npm start
5. Navigate to http://localhost:3000/components/cf-buttons/#button-with-an-animated-icon

## Review

- @caheberer
- @Scotchester 
- @cfarm 
- @contolini 
- @anselmbradford 
- Anyone else that has had/has an opinion about icons and atomic organization

## Screenshots

screen shot 2017-04-03 at 4 47 08 pm

## Notes

- As much as I dislike using an empty element, and I was part of the push to remove them for icons, this is a bit of a complex situation. @Scotchester and I discussed the option to remove the border for animated icons and offsetting the rotation to account for the padding to avoid using an empty span. While short term I agree this keep things tidy and keeps from having a breaking change, long term it restricts the possibilities for animating the icon, especially when we move to svg icons and the possibilities for animation are opened up.
- This also brings up a good question of should we be animating icons within buttons and what are the use cases beyond loading data.
- See https://github.com/cfpb/design-manual/issues/227 for back discussion.

## Todos

- I didn't include a changelog entry because this is technically a breaking change. At the same time, none of v4 has been implemented yet, so I don't know that there's a risk of marking it as a minor change.
- Discuss the speed of the spin animation, it's really slow. @caheberer, do you have input on this?
- Move to SVG icons. If you look closely you can see the icon wobbles a bit due to the fact that it's a font icon and has a slightly offset baseline.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ are 100% width when stacked on small screens
 
 ### Changed
 - **capital-framework:** [PATCH] Updated Travis script to print which components are published to npm
+- **cf-buttons:** [MINOR] Fixed a bug when animating icons within a button. Changes button icon markup,
+  see usage doc for details
 
 ### Removed
 -
--- a/src/cf-buttons/src/atoms/buttons-with-icons.less
+++ b/src/cf-buttons/src/atoms/buttons-with-icons.less
@@ -2,31 +2,30 @@
 
 // Icon locations
 
-.a-btn__icon-on-left:before {
+.a-btn_icon__on-left {
     padding-right: unit( 10.5px / @btn-font-size, em );
     border-right: 1px solid mix( @btn-bg, @btn-text, 50% );
     margin-right: unit( 7px / @btn-font-size, em );
 }
 
-.a-btn__icon-on-right:after  {
+.a-btn_icon__on-right {
     padding-left: unit( 10.5px / @btn-font-size, em );
     border-left: 1px solid mix( @btn-bg, @btn-text, 50% );
     margin-left: unit( 7px / @btn-font-size, em );
 }
 
 
-.a-btn__icon-on-left:before,
-.a-btn__icon-on-right:after {
-    .a-btn__secondary& {
+.a-btn_icon {
+    .a-btn__secondary & {
         border-color: mix( @btn__secondary-bg, @btn__secondary-text, 50% );
     }
 
-    .a-btn__warning& {
+    .a-btn__warning & {
         border-color: mix( @btn__warning-bg, @btn__warning-text, 50% );
     }
 
-    .a-btn__disabled&,
-    .a-btn[disabled]& {
+    .a-btn__disabled &,
+    .a-btn[disabled] & {
         border-color: mix( @btn__disabled-bg, @btn__disabled-text, 50% );
     }
 }
--- a/src/cf-buttons/usage.md
+++ b/src/cf-buttons/usage.md
@@ -583,7 +583,7 @@ For accessibility reasons, use the semantic `<button>` instead of a link when po
 
 #### Button with an animated icon
 
-<button class="a-btn">
+<button class="a-btn a-btn__disabled">
     Submit your complaint
     <span class="a-btn_icon
                  a-btn_icon__on-right
--- a/src/cf-buttons/usage.md
+++ b/src/cf-buttons/usage.md
@@ -431,150 +431,178 @@ For accessibility reasons, use the semantic `<button>` instead of a link when po
 
 #### Button with icon on the left
 
-<button class="a-btn
-               a-btn__icon-on-left
-               cf-icon
-               cf-icon__before
-               cf-icon-delete">
-  Close
+<button class="a-btn">
+    <span class="a-btn_icon
+                 a-btn_icon__on-left
+                 cf-icon
+                 cf-icon__before
+                 cf-icon-delete"></span>
+    Close
 </button>
 
-<button class="a-btn
-               a-btn__icon-on-left
-               a-btn__secondary
-               cf-icon
-               cf-icon__before
-               cf-icon-delete">
-  Close
+<button class="a-btn a-btn__secondary">
+    <span class="a-btn_icon
+                 a-btn_icon__on-left
+                 cf-icon
+                 cf-icon__before
+                 cf-icon-delete"></span>
+    Close
 </button> - Secondary button
 
-<button class="a-btn
-               a-btn__icon-on-left
-               a-btn__warning
-               cf-icon
-               cf-icon__before
-               cf-icon-delete">
-  Close
+<button class="a-btn a-btn__warning">
+    <span class="a-btn_icon
+                 a-btn_icon__on-left
+                 cf-icon
+                 cf-icon__before
+                 cf-icon-delete"></span>
+    Close
 </button> - Warning button
 
-<button class="a-btn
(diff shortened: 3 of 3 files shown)


marteki:
> This also brings up a good question of should we be animating icons within buttons and what are the use cases beyond loading data.

For the Before You Claim project, we felt keeping the icon close to or within the button was important so people could know that the page had changed and their interaction was registered. Once the button is clicked, the tool makes calls to SSA's website; it usually takes less than a second on a high-speed connection, but sometimes takes as long as five or six seconds. The length of time might be even longer for folks on slower connections. The icon is serving as an indicator to the user that something is happening in the background, even if it takes ten seconds.

There's a bit more info on our internal GitHub, on the `bugs-and-issues` repo for this tool. Search for "spinner" and look for a closed issue.

> Discuss the speed of the spin animation, it's really slow. @caheberer, do you have input on this?

We leaned towards slower partially to not unknowing trigger any neurovestibular disorders that are sensitive to spinning, and partially to counteract the potential long length of time the spinner would be going.

jimmynotjim:
> we felt keeping the icon close to or within the button was important so people could know that the page had changed and their interaction was registered

I had assumed this was the reason but thank you for verifying it. We should include something about this in the button specific content in the DM.

> We leaned towards slower partially to not unknowing trigger any neurovestibular disorders that are sensitive to spinning, and partially to counteract the potential long length of time the spinner would be going.

Ah, that's a good use case, we should include something about that in the animation content in the Design Manual. I'll be sure to ping you when I open that PR for feedback.

Realized I should have included this for anyone that doesn't know what the icon spin speed is currently https://cfpb.github.io/capital-framework/components/cf-icons/#animated-icons.

Scotchester:
> we felt keeping the icon close to or within the button was important

Do you think that having a state icon (loading spinner, confirmation check, etc.) _next_ to a button would be equally as effective as _within_ the button?

caheberer:
> Do you think that having a state icon (loading spinner, confirmation check, etc.) next to a button would be equally as effective as within the button?

I know we explored that as an option when designing this solution. If I remember correctly, we were worried about visibility, that folks might still miss it and also then needing to show the button in an inactive state (or visually different) while the loading process was happening. @Scotchester is <i>next</i> to a more desired solution for you? If so, just curious to hear your rationale. 

Separate but similar issue we might want to add (because we use the animation in a button and in a page-level alert):
Make a distinction that when a button loads content within the same page, an icon inside a button is used, and when a button loads a new page, a loading icon within a page-level alert is used.

jimmynotjim:
> also then needing to show the button in an inactive state

Now I wonder if we should do this even with the icon inside the button so it's clear that you can't keep clicking the button to make it go faster (especially since that'll probably have the inverse affect)

Scotchester:
> Now I wonder if we should do this even with the icon inside the button so it's clear that you can't keep clicking the button to make it go faster (especially since that'll probably have the inverse affect).

The JS that changes the class to make the icon change/spin can also disable the button.

Scotchester:
> @Scotchester is next to a more desired solution for you? If so, just curious to hear your rationale.

It is. If we all agreed that putting it next to the button was fine for the UX, then we wouldn't have to worry about animating icons within buttons, and this whole PR and its attendant complications for CF versioning would be unnecessary.

jimmynotjim:
> The JS that changes the class to make the icon change/spin can also disable the button.

Right, but we don't have that documented and it's not how I built the example. Both should probably happen, but I'd definitely want help from the design side on that.

jimmynotjim:
Looking at our typical form example, I don't know that adding a current state icon between the two buttons is a great option, it feels very awkward and it's not clear which button it's attached to, where as the icon within the button looks far more refined.

screen shot 2017-04-05 at 1 25 21 pm

screen shot 2017-04-05 at 1 28 02 pm

screen shot 2017-04-05 at 1 29 26 pm 

(ignore the border color, I added the styles inline to make this one work from the inspector)

caheberer:
Thanks @jimmynotjim for your latest thoughts. Now that you've posted it here, I remember our team encountering that exact issue. For this reason, I think inside the button is better from a ux perspective @marteki agree?

It's also the most common convention I've seen outside the Bureau.

marteki:
Yeah, I also agree. We'd want to make it really clear which action the spinner was related to. Having the spinner appear as a part of the action button meets that need, and doesn't introduce as much uncertainty about what the UI is trying to say to the user.

Also: on the Retirement example shown, I believe we either set the button inactive or rejected further responses on the backend while the API call was still ongoing. We could have a more explicit indicator on the UI side. ("DON'T CLICK THIS ANYMORE, IT'S TRYING TO DO THE THING FOR YOU" but obviously not just saying that.) Do users need a more explicit indicator?

jimmynotjim:
I think making it inactive with the animation is plenty of hints to the user. I've made that update in the doc example.

screen shot 2017-04-06 at 11 27 16 am

Scotchester:
@jimmynotjim Your code example doesn't show the `a-btn__disabled` modifier.

jimmynotjim:
@Scotchester I guess I should push that eh?

jimmynotjim:
So it sounds like a consensus about the icon staying inside the button. @Scotchester do you have any objections or can we move forward here?

Scotchester:
I'd still rather narrow the exception to standard icon handling only to _animated_ icons within buttons, and leave standard buttons with static icons alone. Perhaps we should get another FEWD or two to weigh in on that issue.

Whichever way we go on that, as long as we agree it's not necessary to bump to 5.0.0 for this, I'm good with it. (Don't forget to update the change log when we have consensus on that point.)

jimmynotjim:
One last possible scenario, say we start with a static icon in a button, like the save icon, then switch to an animated update icon, then a static approved or error icon.

Why change the markup simply because it's animated in one of those three states when we could just change the class names? It's far simpler to maintain future code with the suggestions laid forward here.

ascott1:
@Scotchester and @jimmynotjim it seems like there is a disagreement on implementation, but this looks like a reasonable solution. While I there is certainly merit to both sides of the discussion, I like the simplicity of having a single pattern for buttons, rather than having to document exceptions or modify markup with JS. I'd recommend merging this, but creating a new CF issue documenting this as less than ideal. That moves us forward while leaving it open to future improvement should we have an idea for a different implementation.

What do you two think?

Scotchester:
Jimmy's final example scenario was enough to convince me, for now, I just forgot to post that.

I agree with merging, again provided we call this 4.2.0 or whatever it is, not 5.0.0.

ascott1:
> 4.2.0 or whatever it is, not 5.0.0.

👍

PR #469, opened 2017-04-14
Update Capital Framework to include the latest hero updates to the Design Manual

jimmynotjim:
The documentation in the Design Manual has been updated to include additional guidelines for heros. This change includes updated rules about image sizes and spacing. The changes to Capital Framework include an attempt to simplify the markup and offer additional image sizes through inline styles.

## Additions

- Transitioned from a single inline background image to `image-set` (with backups) to allow modern devices to swap the image used depending on the device
- Added the bleeding hero markup example to the docs

## Changes

- Reduced the hero markup to a single markup pattern that works across all the modifiers
- Switched from table-cell to flex to vertically center the content
- Switched from static height image to use the flexible container mixin
- Changed the `overlay-text` modifier to `knockout` to better describe the modification based on feedback from @nataliafitzgerald
- Removed the duplicate image block for the bleeding hero
- Fixed the grid wrapper width

## Removed

- Removed the CTA example and supporting styles because the hero pattern no longer allows for CTAs

## Testing

- Due to the full width nature of the hero, it's better to test it on the sandbox rather than using the docs site. To do this:

1. checkout this branch locally and run `npm run cf-link`
2. checkout the `hero-updates` branch of capital-framework-sandbox locally (if you haven't before, run `bundle install && npm install`
3. in the sandbox run `npm run cf-link && npm run build && npm start`
4. navigate to `http://localhost:3000/components/cf-layout/heroes.html`

## Review

- @virginiacc 
- @nataliafitzgerald
- @Scotchester 
- @anselmbradford 
- @sebworks 

## Screenshots

screen shot 2017-04-13 at 7 01 42 pm

screen shot 2017-04-13 at 7 01 55 pm

screen shot 2017-04-13 at 7 01 22 pm

screen shot 2017-04-13 at 7 02 20 pm

## Notes

- https://github.com/cfpb/design-manual/issues/461

## Checklist

### DIFF
--- a/src/cf-layout/src/molecules/heroes.less
+++ b/src/cf-layout/src/molecules/heroes.less
@@ -5,37 +5,38 @@
 // Sizing variables
 
 @hero-desktop-height: 285px;
-@hero-image-height:   195px;
 
 // Color variables
 // $color- variables are from 18F's U.S. Web Design Standards
 // https://github.com/18F/web-design-standards/blob/master/src/stylesheets/core/_variables.scss
 
 @hero-bg:             #dce4ef; // $color-gray-cool-light
-@hero-overlay-bg:     #5b616b; // $color-gray
-@hero-overlay-text:   white;
+@hero-knockout-bg:    #5b616b; // $color-gray
+@hero-knockout-text:  white;
 
 
 .m-hero {
     background-color: @hero-bg;
-
-    .respond-to-min( @bp-sm-min, {
-        display: table;
-        width: 100%;
-        height: @hero-desktop-height;
-    } );
 
     &_wrapper {
         padding-top: unit( @grid_gutter-width / @base-font-size-px, em );
         padding-bottom: unit( @grid_gutter-width / @base-font-size-px, em );
 
         .respond-to-min( @bp-sm-min, {
-            display: table-row;
+            // IE specific fix
+            display: table;
+
+            @supports( display: flex ) {
+                display: flex;
+            }
+
+            min-height: @hero-desktop-height - ( @grid_gutter-width * 2 );
         } );
 
-        .respond-to-max( @bp-xs-max, {
-            padding-top: unit( @grid_gutter-width / @base-font-size-px, em );
-            padding-bottom: unit( @grid_gutter-width / @base-font-size-px, em );
+        .respond-to-min( @bp-lg-min, {
+            padding-top: unit( ( @grid_gutter-width * 1.5 ) / @base-font-size-px, em );
+            padding-bottom: unit( ( @grid_gutter-width * 1.5 ) / @base-font-size-px, em );
+            min-height: @hero-desktop-height - ( ( @grid_gutter-width * 1.5 ) * 2 );
         } );
     }
 
@@ -45,12 +46,14 @@
         .respond-to-min( @bp-sm-min, {
             .grid_column( 7, 12 );
 
-            padding: unit( @grid_gutter-width / @base-font-size-px, em )
-                     unit( 15px / @base-font-size-px, em );
-
-            // Keep the text vertically centered
+            // IE specific fix
             display: table-cell;
             vertical-align: middle;
+
+            @supports ( display: flex ) {
+                display: block;
+                margin: auto;
+            }
         } );
     }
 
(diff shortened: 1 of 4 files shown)


jimmynotjim:
Part of this PR changes the suggested markup pattern to use `image-set` to allow for responsive background images. At this moment, that's not really possible from within the CMS. That said, the default backup background image will work just fine if we continue serving @2x images. It's not ideal for devices that don't need it, but it's a smaller backend lift to change from individual inline styles to a `style` element that collects each of them as one bundle.

src/cf-layout/src/molecules/heroes.less
```diff
@@ -45,12 +40,11 @@
...
             .grid_column( 7, 12 );
 
-            padding: unit( @grid_gutter-width / @base-font-size-px, em )
-                     unit( 15px / @base-font-size-px, em );
+            margin: auto;
 
             // Keep the text vertically centered
-            display: table-cell;
-            vertical-align: middle;
+            //display: table-cell;
```
jimmynotjim: I still need to do some testing in IE to see whether we need these as a backup and if so, how best to deliver them just in those situations.

src/cf-layout/src/molecules/heroes.less
```diff
@@ -62,133 +66,121 @@
...
-        }
-    }
+            }
 
-    &_image {
-        // height is the same as min-height in a table
-        height: unit( @hero-image-height / @base-font-size-px, em );
+            .m-hero_image {
+                // Remove the flexible container padding so that the image is a consistent width
+                padding-bottom: 0!important;
```
anselmbradford: Add space `0 !imporant;`

src/cf-layout/src/molecules/heroes.less
```diff
@@ -62,133 +66,121 @@
...
 
     &_image-wrapper {
         .grid_column( 1, 1 );
 
+        overflow: hidden;
+
         .respond-to-min( @bp-sm-min, {
             .grid_column( 5, 12 );
 
-            // Keep the image vertically centered
```
anselmbradford: Is this comment still applicable to `margin:auto` or no?

src/cf-layout/usage.md
```diff
@@ -1491,54 +1491,352 @@ height. The image should be `195px` in height to conform to this standard.
...
     <div class="m-hero_wrapper wrapper">
         <div class="m-hero_text">
-            <h1 class="m-hero_heading">Hero title</h1>
+            <h1 class="m-hero_heading">Standard hero with illustration</h1>
             <p class="m-hero_subhead">
-                Hero text goes here. This paragraph has a recommended maximum
-                length of 185 characters. This paragraph has a recommended
-                maximum length of 185 characters.
+                This text has a recommended count of 165-186 characters
+                (three lines at 1260px) following a one-line heading
```
anselmbradford: Is this `1260` significant? Should it be `1230` per the standard width?

src/cf-layout/src/molecules/heroes.less
```diff
@@ -62,133 +66,121 @@
...
+            .m-hero_image-wrapper {
+                margin-top: unit( @grid_gutter-width / @base-font-size-px, em ) * -1;
+                margin-bottom: unit( @grid_gutter-width / @base-font-size-px, em ) * -1;
 
-            .respond-to-min( @bp-sm-min, {
-                .m-hero_image {
-                    display: none;
-                }
-            } );
+                //IE specific fix
```
jimmynotjim: It's all versions, no versions of IE support Flex correctly.

anselmbradford: IE10 and 11 do (with bugs)

anselmbradford: So, in this case, they won't be using `position: absolute` I imagine, as I would think `@supports ( display: flex ) {` would evaluate to true for IE10/11. That's my guess anyway.

src/cf-layout/src/molecules/heroes.less
jimmynotjim: It's still applicable to everything from L:90-L:98, but I don't know that it's all that useful. The spec calls out the vertically centered alignment. If you think it still is I can put it back.

src/cf-layout/src/molecules/heroes.less
jimmynotjim: IE10 and 11 do, like you said, with bugs (and it's serious bugs) but they don't report that they do through `@supports`, so the table cell alignment is for all IE browsers and whatever webkit ones that don't report support for flex. As of now, the latest builds of Chrome, Safari, and FF are the only ones reporting through `@supports` as actually supporting Flex

src/cf-layout/src/molecules/heroes.less
```diff
@@ -5,37 +5,38 @@
...
+            min-height: @hero-desktop-height - ( @grid_gutter-width * 2 );
         } );
 
-        .respond-to-max( @bp-xs-max, {
-            padding-top: unit( @grid_gutter-width / @base-font-size-px, em );
-            padding-bottom: unit( @grid_gutter-width / @base-font-size-px, em );
+        .respond-to-min( @bp-lg-min, {
+            padding-top: unit( ( @grid_gutter-width * 1.5 ) / @base-font-size-px, em );
+            padding-bottom: unit( ( @grid_gutter-width * 1.5 ) / @base-font-size-px, em );
+            min-height: @hero-desktop-height - ( 45px * 2 );
```
jimmynotjim: Fixed

src/cf-layout/src/molecules/heroes.less
jimmynotjim: Fixed

src/cf-layout/usage.md
jimmynotjim: Yep, no idea how I messed that up. Fixed.

src/cf-layout/src/molecules/heroes.less
```diff
@@ -62,133 +65,121 @@
...
-        }
-    }
+            }
 
-    &_image {
-        // height is the same as min-height in a table
-        height: unit( @hero-image-height / @base-font-size-px, em );
+            .m-hero_image {
+                // Remove the flexible container padding so that the image is a consistent width
+                padding-bottom: 0 !important;
```
sebworks: why is important necessary?

jimmynotjim: Discussed in chat. Because the images are being loaded through inline styles, the padding must be recalculated if it's not a standard sized image, and that inline style has a higher specificity. 

I had originally absolutely positioned the flexible container within the image wrapper, but that proved far more complex due to the height of the image changing as you reduced the browser. So something like this didn't work unless the browser was at full width:

```
  top: 50%;
  margin: -(half the image height);
```

It turned out to be far simpler to remove the flexible container padding and making the image 100% in height and width of the wrapper.

src/cf-layout/src/molecules/heroes.less
jimmynotjim: screen shot 2017-04-25 at 4 30 16 pm

IE 11 using `display: table` despite supporting the prefixed flexbox when it's outside `@supports`.

jimmynotjim: Turns out IE doesn't support `@supports` at all and most likely never will. I think we just found an IE conditional we can rely on.

PR #484, opened 2017-05-02
Add base styling for inline and fenced code blocks

jimmynotjim:
The project doesn't currently style code blocks in any way. This adds sensible defaults that can easily be overridden through variables.

## Additions

- Added base styling for inline and fenced code blocks

## Testing

- This is difficult to test in any of our doc sites due to the existing code styling that lives in those. 

## Review

- @Scotchester 
- @cfarm 
- @anselmbradford 
- @sebworks 

## Screenshots

screen shot 2017-05-02 at 12 52 34 pm

screen shot 2017-05-02 at 12 58 07 pm

## Checklist

### DIFF
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -539,5 +539,38 @@ figure {
 }
 
 //
+// Code blocks
+//
+
+pre,
+code {
+    background: @code-bg;
+    border-radius: 4px;
+    color: @code-text;
+    font-family: 'Input Mono', Consolas, Monaco, 'Courier New', monospace;
+}
+
+code {
+    display: inline-block;
+    padding: unit( 3px / @size-code, em )
+             unit( 6px / @size-code, em )
+             0;
+    font-size: unit( @size-code / @base-font-size-px, em );
+}
+
+pre {
+    display: block;
+    padding: unit( 10px / @base-font-size-px, em)
+             unit( 15px / @base-font-size-px, em );
+    white-space: pre-wrap;
+    overflow-wrap: break-word;
+
+    code {
+        padding: 0;
+        background-color: transparent;
+    }
+}
+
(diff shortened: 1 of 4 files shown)


Scotchester:
That inline example looks a lot bigger than 13px. Is that actually how it renders out?

src/cf-core/src/cf-base.less
```diff
@@ -539,5 +539,38 @@ figure {
...
+    background: @code-bg;
+    border-radius: 4px;
+    color: @code-text;
+    font-family: 'Input Mono', Consolas, Monaco, 'Courier New', monospace;
+}
+
+code {
+    display: inline-block;
+    padding: unit( 3px / @size-code, em )
+             unit( 6px / @size-code, em )
```
Scotchester: This side padding looks a bit too generous, to my eyes, but maybe it's just me.

src/cf-core/src/cf-base.less
```diff
@@ -539,5 +539,38 @@ figure {
...
+             0;
```
Scotchester: Having 0 bottom padding doesn't feel right. What happens if one of the letters has a descender? Does it get uncomfortably close to the bottom edge?

jimmynotjim:
screen shot 2017-05-16 at 10 40 27 am

@Scotchester I've removed all the custom code styles on the GH pages branch to show what these defaults look like, I think a few were still getting through in the above screenshots.

- The text size looks correct, I'm assuming the previous screenshots were affected by the GH pages branch styles.
- Agreed on the side padding, I've reduced it to 3px to match the top.
- I think the bottom space is ok, the problem with adding bottom padding is it'll mess up the line-height if the inline code has a line of text below it, like in a paragraph.

Scotchester:
:+1: :shipit:

PR #495, opened 2017-05-31
Update contributing doc for testing in the doc site

jimmynotjim:
It's unclear how to test components changes in the documentation site. We've also made testing in a separate project much easier.

## Additions

- Added instructions for testing in the documentation site.

## Changes

- Updated the instructions for testing in an outside project.

## Testing

- Follow the instructions.

## Review

- @anselmbradford 
- @sebworks 
- @cfarm 
- @contolini 
- @Scotchester 

## Notes

- Hopefully this clears some things up for people. Feel free to tear it apart and tell me it's still unclear and I need to make it better. I'm definitely closest to the process at the moment.

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [x] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [x] Passes all existing automated tests
* [x] New functions include new tests
* [x] New functions are documented (with a description, list of inputs, and expected output)
* [x] Placeholder code is flagged
* [x] Visually tested in supported browsers and devices
* [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #498, opened 2017-06-01
Add a level of basic styling to legend elements

jimmynotjim:
Legends are currently unstyled in Capital Framework.

## Additions

- Added legend styling

## Testing

- Follow the [instructions to test the CF Docs](https://github.com/cfpb/capital-framework/pull/495/files#diff-6a3371457528722a734f3c51d9238c13)
- Navigate to `http://localhost:3000/components/cf-forms/#legends`

## Review

- @virginiacc 
- @Scotchester 
- @anselmbradford 

## Screenshots

screen shot 2017-06-01 at 3 57 16 pm

## Notes

- See https://github.com/cfpb/cfgov-refresh/commit/1412edfb93c8b6d4b0ca0e1e9fe54adc21a599c8#diff-bb9725e42c42447534e5b670dbe94b8c for source

## Todos

- Note legend styling in cfgov-refresh has been ported.

## Checklist

### DIFF
--- a/src/cf-forms/src/atoms/legend.less
+++ b/src/cf-forms/src/atoms/legend.less
@@ -1,0 +1,10 @@
+.a-legend {
+    .h4();
+
+    // Legends do not wrap in IE.
+    // Different styles are required to ensure wrapping in different versions.
+    box-sizing: border-box; // IE9-11 & Edge 12-13
+    display: table; // IE8-11
+    max-width: 100%; // Patch for IE9-11 & Edge 12-13
+    white-space: normal; // IE8-11
+}
--- a/src/cf-forms/usage.md
+++ b/src/cf-forms/usage.md
@@ -88,14 +88,14 @@ Overwrite them in your own project by duplicating the variable `@key: value`.
 
 ## Legends
 
-<label class="a-legend">
+<legend class="a-legend">
     A basic legend
-</label>
+</legend>
 
 ```
-<label class="a-legend">
+<legend class="a-legend">
     A basic legend
-</label>
+</legend>
 ```
 
 ## Labels
--- a/src/cf-forms/usage.md
+++ b/src/cf-forms/usage.md
@@ -15,6 +15,7 @@ Capital Framework.
 
 - [Variables](#variables)
     - [Color variables](#color-variables)
+- [Legends](#legends)
 - [Labels](#labels)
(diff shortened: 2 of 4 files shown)


Scotchester:
Shouldn't this just be applied to the `legend` element in `cf-base.less`? I don't see the need for considering this an atom.

jimmynotjim:
@Scotchester we moved all the form items out of cf-base during the v4 work because it was confusing that half was in there and half in cf-forms.

src/cf-forms/src/atoms/legend.less
```diff
@@ -0,0 +1,10 @@
+.a-legend {
+    .h4();
+
+    // Legends do not wrap in IE.
+    // Different styles are required to ensure wrapping in different versions.
+    box-sizing: border-box; // IE9-11 & Edge 12-13
+    display: table; // IE8-11
```
anselmbradford: This is probably fine, but elsewhere we have things like this wrapped in .lt-ie classes. The one thing would be if we dropped support for a particular IE we could search for the classes associated with that version.

Scotchester:
@jimmynotjim Ugh, right. "Confusing." And applying the same class to every instance of a standard HTML element isn't confusing at all... :fist_right: 🐴 💀

src/cf-forms/usage.md
```diff
@@ -85,6 +86,17 @@ Overwrite them in your own project by duplicating the variable `@key: value`.
...
+## Legends
+
+<label class="a-legend">
+    A basic legend
+</label>
+
+```
+<label class="a-legend">
+    A basic legend
+</label>
```
Scotchester: Is there a reason why the HTML element used here isn't `legend`?

jimmynotjim: Because I'm a dummy. I'll fix it.

jimmynotjim:
@Scotchester fixed.

src/cf-forms/src/atoms/legend.less
Scotchester: Didn't IE drop conditional comments at some point? I don't think we can do the class-based thing for all of these cases.

jimmynotjim: Yeah, IE10 and 11 are not targetable with conditional comments and from what I've read, it's a messy road trying to make it happen.

PR #513, opened 2017-06-15
Fixes pagination docs

anselmbradford:
## Changes

- Updates outdated class name references for `a-btn__icon-on-left` and `a-btn__icon-on-right`, which have updated names in https://github.com/cfpb/capital-framework/blob/canary/src/cf-buttons/src/atoms/buttons-with-icons.less#L15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
 ### Changed
 - **cf-forms:** [PATCH] Fixed checkbox/radio borders on hover
+- **cf-pagination:** [PATCH] Fixed outdated `a-btn__icon-on-left`
+and `a-btn__icon-on-right` classes referenced in docs to correct
+`a-btn_icon__on-left` and `a-btn_icon__on-right` classes.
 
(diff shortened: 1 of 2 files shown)


Scotchester:
:+1:

Scotchester:
We'll need a change log entry if we want a new patch release for it.

anselmbradford:
> We'll need a change log entry if we want a new patch release for it.

Added to `CHANGELOG.md`! Do I need to do anything else?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
 ### Changed
 - **cf-forms:** [PATCH] Fixed checkbox/radio borders on hover
+- **cf-pagination:** [PATCH] Fixed outdated `a-btn__icon-on-left`
+and `a-btn__icon-on-right` classes referenced in docs.
```
Scotchester: This is nitpicky, but I think the change log should probably give either _both_ old and new class names, or just the correct ones.

anselmbradford: Okay, added both!

PR #514, opened 2017-06-20
Remove IE7 support

anselmbradford:
Folks, is it time???

## Removals

- Removes references to `lt-ie8` class and associated style rules that are used only for IE7 and below support.

## Testing

- Should not work as expected in IE7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and `a-btn__icon-on-right` classes referenced in docs to correct
 `a-btn_icon__on-left` and `a-btn_icon__on-right` classes.
 
 ### Removed
--
+- **capital-framework:** [MINOR] Remove IE7 and below support.
 
 ## 4.7.1 - 2017-06-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@ and `a-btn__icon-on-right` classes referenced in docs to correct
 `a-btn_icon__on-left` and `a-btn_icon__on-right` classes.
 
 ### Removed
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -11,7 +11,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 - **cf-forms:** [PATCH] Fixed checkbox/radio borders on hover
 
 ### Removed
--
+- **capital-framework:** Remove IE7 and below support.
```
jimmynotjim: This should probably be a `[MINOR]` right? It doesn't really change much but feels more than a PATCH

anselmbradford: Fixed!

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ and `a-btn__icon-on-right` classes referenced in docs to correct
 `a-btn_icon__on-left` and `a-btn_icon__on-right` classes.
 
 ### Removed
--
+- **capital-framework:** [MINOR] Remove IE7 and below support.
```
Scotchester: I'm pretty sure we need to list the modified components individually to ensure they get bumped. Is that right, @contolini?

jimmynotjim: Oh right, I always forget what can and can't be bumped with the entire project.

anselmbradford: Broken into individual modules in https://github.com/cfpb/capital-framework/pull/514/commits/01403fa2de19eaa100e0e6f7dcaf62864e7679ab

PR #519, opened 2017-06-26
Updates heading level to match design manual

anselmbradford:
## Changes

- Updates heading level of labels to match design manual https://cfpb.github.io/design-manual/page-components/form-fields.html#labels-and-helper-text

## Testing

- Follow instructions in https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#changing-the-codebase

## Notes

- The margin override was a hardcoded 10px originally, but is ems in cfgov refresh so updating here. I think that's correct?

### DIFF
--- a/src/cf-forms/src/atoms/label.less
+++ b/src/cf-forms/src/atoms/label.less
@@ -3,10 +3,10 @@
 }
 
 .a-label__heading {
-    .heading-5();
+    .heading-4();
 
     display: block;
 
-    // Overwrites heading-5 margin.
-    margin-bottom: 10px;
+    // Overwrites heading-4 margin.
+    margin-bottom: unit(10px / @font-size, em);
 }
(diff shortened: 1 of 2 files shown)


src/cf-forms/src/atoms/label.less
```diff
@@ -3,10 +3,10 @@
...
 .a-label__heading {
-    .heading-5();
+    .heading-4();
 
     display: block;
 
-    // Overwrites heading-5 margin.
-    margin-bottom: 10px;
+    // Overwrites heading-4 margin.
+    margin-bottom: unit(10px / @font-size, em);
```
jimmynotjim: Minor: be sure to add spaces within the parens

anselmbradford: Fixed in https://github.com/cfpb/capital-framework/pull/519/commits/4cb65cb2a0a4b59cb70fce32bd08f90e6626ee38

PR #521, opened 2017-06-27
Updates style deps and removes unused deps

anselmbradford:
## Removes
- Removes unused `less-plugin-npm-import`, `inquirer`, `versiony`.

## Changes

- Updates `gulp-autoprefixer` to `4.0.0`.
- Updates `gulp-uglify` to `3.0.0`.
- Updates minor version of `gulp-cssmin` and `gulp-less` dependencies, but these have a `^` so this shouldn't functionally change anything

## Testing

- `npm install && npm run build` should pass.


CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and `a-btn__icon-on-right` classes referenced in docs to correct
 `a-btn_icon__on-left` and `a-btn_icon__on-right` classes.
 - **cf-forms:** [PATCH] Changed label heading from 5 to 4, per design manual.
 - **cf-typography:** [PATCH] Remove bottom margin from `m-slug-header`.
+- **capital-framework:** Updates `gulp-autoprefixer` to `4.0.0`.
+- **capital-framework:** Updates `gulp-uglify` to `3.0.0`.
```
Scotchester: I'm unclear on whether we need a scope of change designation here. @contolini?

anselmbradford:
Removes `inquirer` too

CHANGELOG.md
anselmbradford: Fixed in https://github.com/cfpb/capital-framework/pull/521/commits/fc4abaa85862ed478eedd36033e1556cc11775a1

contolini: It doesn't need it but to avoid confusion I vote we agree to *always* add one.

Under the hood, if `capital-framework` doesn't have a semver version (major, minor, patch) specified it'll use the component's semver version.

It's possible the bot will erroneously double increment `capital-framework` because we're listing it twice. It's okay if that happens, we can fix it and write a regression test. It's good to push the limits of the bot.

anselmbradford:
Squashed and updated to one commit.

PR #528, opened 2017-06-30
Convert form units to ems

anselmbradford:
## Changes

- Convert fixed pixel units to ems in the forms.

## Testing

- Check out this branch and follow https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-components-locally
- Check cf-forms.

### DIFF
--- a/src/cf-forms/src/organisms/form.less
+++ b/src/cf-forms/src/organisms/form.less
@@ -1,6 +1,6 @@
 .o-form {
     &_group {
-        margin-bottom: 30px;
+        margin-bottom: unit( 30px / @base-font-size-px, em );
     }
 
     &_fieldset {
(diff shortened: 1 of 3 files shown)


src/cf-forms/src/organisms/form.less
```diff
@@ -10,7 +10,7 @@
         padding: 0;
 
         .m-field + .m-field {
```
jimmynotjim: Realizing now this is `m-form-field` not `m-field`

anselmbradford: Fixed in https://github.com/cfpb/capital-framework/pull/528/commits/9f7e7bd4d6498c2ae581a3e5e1efe5dd07be7d84

anselmbradford:
Ready for review again!

PR #538, opened 2017-07-11
Adds expandable intro content example to docs

anselmbradford:
## Additions

- Adds expandable group intro content defined in https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/expandable-group.html#L30-L40 to the docs.

## Removals

- Removed unused reference to data- attribute for accordion docs.

## Testing

- Follow https://cfpb.github.io/capital-framework/contributing/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,15 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 ## Unreleased
 
 ### Added
-- 
+-
 
 ### Changed
 - **cf-forms:** [PATCH] Convert fixed pixel units to ems.
+- **cf-expandables:** [PATCH] Added example in docs for intro content.
+Removed unused data-attribute from docs.
 
 ### Removed
-- 
+-
 
 ## 4.8.2 - 2017-07-03
 
--- a/src/cf-expandables/usage.md
+++ b/src/cf-expandables/usage.md
@@ -30,7 +30,8 @@ dependencies of this component.
   - [Default state - open on load](#default-state-open-on-load)
   - [Barebones example](#barebones-expandable)
 - [Expandable groups](#expandable-groups)
-  - [Accordion style group](#accordion-style-group)
+  - [Intro-style expandable group](#intro-style-expandable-group)
+  - [Accordion-style expandable group](#accordion-style-expandable-group)
 
 
 ## Variables
@@ -516,13 +517,21 @@ In this barebones example there are no visual styles.
 </div>
 ```
 
-### Accordion-style group
+### Intro-style expandable group
 
-Accordions can only show one open expandable at a time.
-Add the `data-accordion="true"` attribute to the expandable group to activate
-the accordion mode.
+Expandable groups can contain a top horizontal rule,
+a heading, and an introductory paragraph.
 
-<div class="o-expandable-group o-expandable-group__accordion" >
+<div class="o-expandable-group">
+    <div class="a-rule-break"></div>
+    <h3>Expandable Group Header</h3>
+    <p>
+        Lorem ipsum dolor sit amet, consectetur adipisicing
+        elit. Neque ipsa voluptatibus soluta nobis unde quisquam
+        temporibus magnam debitis quidem. Ducimus ratione
+        corporis nesciunt earum vel est quaerat blanditiis
+        dolore ipsa?
+    </p>
     <div class="o-expandable o-expandable__padded">
         <button class="o-expandable_header o-expandable_target"
                 title="Expand content">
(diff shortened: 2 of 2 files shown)


anselmbradford:
Piiing

Scotchester:
I'm not sure that the intro stuff should be documented here. It's not intrinsic to the expandable group itself, and It has no relation to code that is in Capital Framework. It's just adding a rule/heading/paragraph in front of the actual expandable group, if they're defined in the CMS.

I would suggest removing everything from this PR except for the correction about how to trigger accordion mode and the whitespace fixes.

anselmbradford:
I thought since it's inside the `<div class="o-expandable-group">` it should be included since we could futz around with expandable groups here and break this pattern in cfgov-refresh and not know it because it's not documented. If it was a rule/heading/paragraph adjacent to the expandable group it would be different, but this is inside the expandable group's container, so arguably is part of the pattern.

Scotchester:
This is another good example of where it's somewhat tough to draw the line between CF and and whatever project is using CF. This is only under consideration because of how it's coded in cfgov-refresh to output the intro fields inside the `o-expandable-group` div.

To go back to this point that I made earlier: "it has no relation to code that is in Capital Framework", there is no need for CSS or JS related to the introductory content, so why document it as part of the pattern? Users would expect to be able to just put any ol' heading/paragraph/etc. in front of an Expandable Group and have it work like all other headings/paragraphs on their site. It's not _part of_ the Expandable Group. And it's certainly not a specific style of expandable group, as you're presenting it in these changes. (Accordion groups can have intro text, too.)

My perspective is that the Expandable Group is just the collection of expandables, and whatever might optionally be used to intro it is up to the project that's using an Expandable Group.

All that said, this is not a hill I'm going to die on; it's just a minor documentation quibble.

Scotchester:
Oh, I just thought of a slightly different angle from which to look at it:

The Expandable Group Capital Framework pattern and the Expandable Group Wagtail module are two distinct, but overlapping, concepts.

jimmynotjim:
> This is only under consideration because of how it's coded in cfgov-refresh to output the intro fields inside the o-expandable-group div.

I would have coded this the same way even it weren't in wagtail. The heading/paragraph/etc belong to the group of expandables, so putting it inside the group `div` tells future devs these things are related and somehow different from the content on the rest of the page, don't mess with them without considering that relationship. 

To reiterate a point that I think continues to get lost, we should be encouraging the documentation of these patterns in real world scenarios to give future users and maintainers clues on how they too can use them. One of the common frustrations with framework docs is examples often live in a box and it's not clear how they actually work together.

Scotchester:
> > This is only under consideration because of how it's coded in cfgov-refresh to output the intro fields inside the o-expandable-group div.
>
> I would have coded this the same way even it weren't in wagtail.

Well, OK, I can see why you would do that. But Ans cited the Wagtail implementation in the OP, so you and he may have a slightly different perspective on this. 

> The heading/paragraph/etc belong to the group of expandables

We may just have to agree to disagree about this, but... this concept could be expanded to almost any component that might have a heading or text introducing it. Hypothetically, then, should we do something similar for, say, a table? A form? Maybe those are different because their root organism classes tend to be applied to the `<table>` or `<form>` tags, rather than a wrapper of some kind.

But lets assume I agree to move forward with the basic point of this PR. If the consensus opinion is that a heading, at minimum, and optionally a paragraph (which is actually a full-on RTF in Wagatil, FWIW), is intrinsic to the Expandable Group Pattern , then this update shouldn't pitch it as an alternate style. Those elements should just be added to the base expandable group pattern and to the accordion-style pattern. But as noted earlier, @anselmbradford may have a different perspective on this point.

anselmbradford:
> We may just have to agree to disagree about this, but... this concept could be expanded to almost any component that might have a heading or text introducing it.

I think the difference is this pattern is something that came out of the design for this component (https://[GHE]/flapjack/Modules-V1/wiki/Expandable-Group). It's true that there could be all assortment of combinations that COULD be done (in Wagtail or elsewhere) to bastardize what we have documented here, but what I would like to be documented is the patterns that are sanctioned by the design team. Without that, it becomes unclear when encountering a pattern as to whether it's a rogue hack to make things look a certain way, or if it is actually a pattern we want implementers to be using. 

> Those elements should just be added to the base expandable group pattern and to the accordion-style pattern. But as noted earlier, @anselmbradford may have a different perspective on this point.

That's a good point, the wiki docs always have the intro part, but the DM doesn't, so it should be confirmed that these are two supported patterns (as I suspect it is).

nataliafitzgerald:
@anselmbradford @Scotchester 
I did a quick visual audit of the website and I found three existing use cases for expandable groups on the site (this is just a quick manual audit so I am sure it's not comprehensive):

### 1) Heading (H3) + Expandable Group
Typically a heading followed by a list of things. This approach seems to be by far the most common/widespread.
screen shot 2017-09-29 at 4 45 22 pm

- https://www.consumerfinance.gov/consumer-tools/prepaid-cards/choose-the-right-card/
- https://www.consumerfinance.gov/consumer-tools/getting-an-auto-loan/plan-to-shop-for-your-auto-loan/
- https://www.consumerfinance.gov/owning-a-home/process/prepare/
- https://www.consumerfinance.gov/data-research/cfpb-research-conference/2016-cfpb-research-conference/agenda/
- https://www.consumerfinance.gov/practitioner-resources/library-resources/online-resources/
- https://www.consumerfinance.gov/practitioner-resources/library-resources/program-ideas/
- https://www.consumerfinance.gov/practitioner-resources/library-resources/online-resources/
- https://www.consumerfinance.gov/data-research/cfpb-research-conference/2016-cfpb-research-conference/agenda/

### 2) Heading (H3) + Paragraph + Expandable Group
I only found one instance of this and it doesn't really reflect the use case that I think the original pattern on V1 was meant to address. It's more of a creative use of that pattern than intended use. Which is fine, just worth considering that I couldn't actually find any examples of the pattern being used exactly as it was designed/documented in the V1 atomic wiki. With that being said I can imagine a time where the pattern, as designed, could be useful/needed. 
- https://www.consumerfinance.gov/about-us/the-bureau/leadership-calendar/

### 3) H1 + Lead Paragraph + Expandable Group
Essentially, from a code perspective, you could consider this an Expandable group without a heading or paragraph but the way I see this one from a content perspective is that the H1 and the lead paragraph are standing in as the heading and intro for the Expandable group. 
screen shot 2017-09-29 at 6 07 16 pm

- https://www.consumerfinance.gov/practitioner-resources/your-money-your-goals/online-resources/

Based on this audit, it appears to me that including a heading for an Expandable group is the rule rather than the exception. From a content perspective, I didn't find a single instance without a heading for the Expandable group. The Expandable group should have a heading that captures the bigger idea of why these things are a set (e.g. Action steps, Things that will happen on a day (in the case of the agenda), Resources, Topic areas, etc). Including supporting text after the heading is optional. 

A current limitation of the Expandable options in Wagtail is that a user can't specify the heading level they wish to use. Ideally the heading would follow the heading levels on the page. So if it comes after a H1 section, the user could select H2. A user can do a workaround and add the Heading in the "Body" field. But, the fact that it's a workaround gives the impression to the CMS user that they are doing something that isn't within the realm of what's intended. It feels like a hack and it shouldn't.

From a design perspective, when a user does add a supporting paragraph I think it would make more sense to have 30px of spacing between the bottom of the paragraph and the Expandable group rather than 15px. I don't see this specified one way or the other in the V1 mock-up but visually it looks like that was the design intent. 

**15px (too tight)** 
screen shot 2017-09-29 at 6 02 14 pm

**30px (just right)**
screen shot 2017-09-29 at 6 04 50 pm

On the DM page we focus on the molecule: the Expandable https://cfpb.github.io/design-manual/page-components/expandables.html The Expandable group is a section on the page and we don't mention the heading or the paragraph text. Perhaps it would be helpful, in the page section on Expandable groups, to include a mention and image of a heading and paragraph text so that a user can better understand how to use this pattern. Some content guidelines could be helpful too.

We do mention the different pieces that make up the Expandable group in the V1 Atomic Wiki: https://github.cfpb.gov/flapjack/Modules-V1/wiki/Expandable-Group. I like that the Atomic wiki talks about included atoms and included molecules. It helps you understand pieces make up this pattern. 
screen shot 2017-09-29 at 6 18 56 pm

One question that I am left with, what is the use case for the single Expandable and where are we using it?

anselmbradford:
> One question that I am left with, what is the use case for the single Expandable and where are we using it?

I think it was only used in expandable-like things, such as the beta banner, which aren't truly expandables, so that's moot. But I'm not sure if there are circumstances where a stack of expandables could be made in Wagtail with dynamic content and it may end up theoretically with only one.

jimmynotjim:
When there's one info-unit we still group it, I wonder if we should make grouping expandables, no matter how many a standard.

anselmbradford:
> When there's one info-unit we still group it, I wonder if we should make grouping expandables, no matter how many a standard.

Yes, expandable and expandable group less and JS are already wrapped together in here. Maybe there should be only one organism really, `o-expandable-group`, and individual expandables would be `o-expandable-group_item` or whatever.

jimmynotjim:
> and individual expandables would be o-expandable-group_item or whatever.

I can dig it.

jimmynotjim:
I no longer dig it. Turns out including anything other than the expandables inside the group causes issues with `nth` selectors that the top border of the first expandable depends on. Unfortunately rescinding my 👍

anselmbradford:
> I no longer dig it. Turns out including anything other than the expandables inside the group causes issues with nth selectors that the top border of the first expandable depends on. Unfortunately rescinding my 👍

What does this mean for https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/expandable-group.html#L30-L40  ? Are you addressing that in https://github.com/cfpb/cfgov-refresh/pull/4365 ?

jimmynotjim:
@anselmbradford yep, I moved it outside the group container here https://github.com/cfpb/cfgov-refresh/pull/4365/files#diff-62ffa150f85a8922ad5f88037b0a301bR40. I don't know if this is the best permanent solution, but it fixes the issues I brought up above.

PR #543, opened 2017-07-20
Convert button links to actual buttons

anselmbradford:
## Changes

- Button links were `a` tags, so essentially styling a link to look like a link. This PR converts them to `button`.

## Testing

- Follow instructions in http://localhost:3000/contributing/#testing-components-locally and visit cf-buttons. 

## Screenshots

screen shot 2017-07-20 at 3 23 32 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 -
 
 ### Changed
-- **cf-buttons:** Change button links in docs to actual buttons.
+- **cf-buttons:** [PATCH] Change button links in docs to actual buttons.
 - **cf-forms:** [PATCH] Convert fixed pixel units to ems.
 - **cf-layout:** [PATCH] Fix the background size of overlay hero images
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,13 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 ## Unreleased
 
 ### Added
-- 
+-
 
 ### Changed
-- **cf-buttons:** Change button links in docs to actual buttons.
+- **cf-buttons:** [PATCH] Change button links in docs to actual buttons.
 
 ### Removed
-- 
+-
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,13 +5,13 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 ## Unreleased
 
 ### Added
-- 
+-
 
 ### Changed
-- 
+- **cf-buttons:** Convert button links to actual buttons.
```
jimmynotjim: Super minor: Should this say "Change button links in Docs..." since this isn't changing anything about the cf-buttons code?

Scotchester: Yeah, it should definitely make that clear.

Scotchester:
This makes me realize that this docs page doesn't have an example of a layout where we have a link-style button next to a regular button. (See the first example on [the DM page](http://cfpb.github.io/design-manual/page-components/buttons.html).) As the purpose of the link-style button styles is to align the link-style button with an adjacent regular button, we should be depicting this so it's clear that we don't have any regressions there.

CHANGELOG.md
anselmbradford: I didn't have `[PATCH]` on this either. I added it but now am not sure if it needs it?

Scotchester:
> I didn't have `[PATCH]` on this either. I added it but now am not sure if it needs it?

I would consider this change inconsequential enough that it does not merit bumping and releasing a new version of cf-buttons, so I would remove the `[PATCH]`.

That shouldn't have any ill effects, right @contolini?

contolini:
> I would consider this change inconsequential enough that it does not merit bumping and releasing a new version of cf-buttons, so I would remove the [PATCH].

I agree that it's inconsequential but I'd still vote to bump its version. There's no risk in bumping a version and even a minor documentation improvement like this one is still considered a backwards-compatible bug fix IMHO.

jimmynotjim:
🐢  In order to get docs updates on the site they need to be included in a version update, otherwise they won't be displayed until the next time the package is updated.

PR #544, opened 2017-07-21
Add a full width input modifier to cf-forms

jimmynotjim:
Inputs often need to fill their container, but not always, so it can not be
the default behavior.

## Additions

- Added full width text input modifier

## Testing

- Follow instructions in [http://localhost:3000/contributing/#testing-components-locally](http://localhost:3000/contributing/#testing-components-locally) and visit cf-forms.

## Review

- @Scotchester 
- @anselmbradford 
- @cfarm 
- @sebworks 

## Screenshots

screen shot 2017-07-21 at 9 37 05 am

## Todos

- Update cfgov-refresh feedback forms to use the new modifier.

## Checklist

### DIFF
--- a/src/cf-forms/usage.md
+++ b/src/cf-forms/usage.md
@@ -21,6 +21,7 @@ Capital Framework.
     - [Label heading](#label-heading)
 - [Inputs](#inputs)
     - [Basic Text Inputs](#basic-text-inputs)
+    - [Full-width inputs](#full-width-inputs)
     - [Basic checkboxes](#basic-checkboxes)
     - [Basic radio buttons](#basic-radio-buttons)
     - [Input states](#input-states)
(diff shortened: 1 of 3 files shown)


src/cf-forms/usage.md
```diff
@@ -150,6 +150,34 @@ Inputs should always be paired with a `label` for accessibility reasons.
...
+    <textarea class="a-text-input a-text-input__full"
+              id="textarea-example">Lorem Ipsum</textarea>
+</div>
+
+```
+<div class="m-form-field">
+    <label class="a-label" for="text-input-example">A text input</label>
+    <input class="a-text-input a-text-input__full"
+           type="text"
+           id="text-input example"
```
anselmbradford: missing dash `text-input-example`

jimmynotjim:
@anselmbradford there's no layout in these examples, to reduce the code to just their necessary bits,
 but we've used breaks before so I used them here. Let me know what you think.

screen shot 2017-08-10 at 12 40 36 pm

src/cf-forms/usage.md
```diff
@@ -130,26 +131,54 @@ Inputs should always be paired with a `label` for accessibility reasons.
 
 ### Basic text inputs
 
-<label class="a-label" for="text-input-example">A text input</label>
+<label class="a-label" for="textinput-example">A text input</label>
```
anselmbradford: Should be `textinput-example` and not `text-input-example`? Just making sure there's not an erroneous search and replace at work here.

jimmynotjim: Yes, I realized the `id` on the text area was `textarea` so I updated the inputs to match that example.

anselmbradford: Okay, I think `textarea` is that way because it follows the format of the <textarea> html, which if this followed the html would be two words, but not a biggie.

jimmynotjim: I don't know that it matters all that much, it's only used to make clicking the label focus the input in the docs.

PR #556, opened 2017-08-10
Rebuilding the docs for Capital Framework v4.8.5

jimmynotjim:
Been a bit since we updated the docs site. Includes two dependency updates required for Node 8.


anselmbradford:
Seems fine, this just sucks in all the recent updates, right?

jimmynotjim:
@anselmbradford correct. The only real change I made was updating the two dev deps for node 8.

PR #557, opened 2017-08-10
Adds a-label_helper

anselmbradford:
## Additions

- Adds `a-label_helper` class for setting optional text next to a form input.

## Testing

- Follow contributing docs in CF.

## Screenshots

screen shot 2017-08-10 at 3 39 54 pm

screen shot 2017-08-10 at 3 40 06 pm

screen shot 2017-08-10 at 3 40 12 pm

### DIFF
--- a/src/cf-forms/src/atoms/label.less
+++ b/src/cf-forms/src/atoms/label.less
@@ -3,6 +3,7 @@
 
     &_helper {
         color: @label-helper;
+        font-size: unit( 14px / @font-size, em );
     }
 
     &__heading {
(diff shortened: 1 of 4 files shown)


src/cf-forms/src/atoms/label.less
```diff
@@ -1,12 +1,16 @@
 .a-label {
     display: inline-block;
-}
 
-.a-label__heading {
-    .heading-4();
+    &_helper {
+        color: @label-helper;
```
jimmynotjim: This should also be 14px text right?

anselmbradford: It should be wrapped in a <small> I think?

anselmbradford: I guess it can be both.

anselmbradford: Done

anselmbradford:
Please check again

Issue #440, opened 2017-01-13
Add notifications component


jimmynotjim:
Notifications have been [added to the Design Manual](https://github.com/cfpb/design-manual/pull/366). Post v4, add a cf-notifications component and update the Design Manual when it's ready.

cfarm:
notification element can be ported from cfgov-refresh: https://github.com/cfpb/cfgov-refresh/search?utf8=%E2%9C%93&q=notification&type=

cfarm:
Ans already got started https://github.com/cfpb/capital-framework/pull/560

PR #560, opened 2017-08-15
Add cf-notification

anselmbradford:
Fixes https://github.com/cfpb/capital-framework/issues/440

## Additions

- Adds cf-notification 1.0.

## Testing

- Pull this down, run `npm run build`
- `cd tmp/cf-notification`
- `npm link`
- Change to another CF repo and `git checkout gh-pages-notifications`.
- Run `npm link cf-notification`
- Check http://localhost:3000/components/cf-notification/

## Screenshots

screen shot 2017-08-14 at 8 57 43 pm
screen shot 2017-08-14 at 8 57 52 pm
screen shot 2017-08-14 at 8 57 57 pm


src/cf-notification/src/cf-notification.less
```diff
@@ -0,0 +1,44 @@
+/* ==========================================================================
+   Capital Framework
+   Notifications Styling
+   ========================================================================== */
+
+//
+// Theme variables
+//
+
+// Color variables
```
anselmbradford: I didn't [include the 18F reference](https://github.com/cfpb/capital-framework/blob/canary/src/cf-pagination/src/cf-pagination.less#L12) (which is also a dead link and needs an update) as I wasn't sure what we're doing with this going forward? How should I handled this?

src/cf-notification/src/cf-notification.less
```diff
@@ -0,0 +1,44 @@
...
+
+@notification-bg:             #d2d3d5;
```
anselmbradford: These are `@m-notification-…` in cfgov-refresh. Do we want the atomic prefix on variables?

anselmbradford:
Note to self: add `screenshot.png`.

anselmbradford:
We may want to combine this with cf-forms instead. Thoughts?

nataliafitzgerald:
@anselmbradford 
Ok, thanks for setting me up to view this locally. It really helps to see it in live code. 

**Action**
- Background: Change to Gray 10 (#f7f8f9)
- Border: Change to Gray 40 (#B4B5B6)

Here's a mock up:
screen shot 2017-08-18 at 5 24 03 pm

**Explanation**
- Reduce space below "m-notification-message" to 5px (it is currently set to the default for space under an H4 which is 15px. This just looks like a lot in this use case)
- Change "m-notification-explanation" to standard paragraph text: Avenir Next Regular, 16px.
screen shot 2017-08-18 at 5 56 54 pm

**Custom icons**
- Match to styling shown above for "Action"

**Icon**
- Can we center the icon vertically so that it is centered with the "m-notification-message"? Right now it looks like it's top aligned with the text.
- The icons are getting cut off on the left side 
screen shot 2017-08-18 at 5 59 17 pm

Scotchester:
Can't give this a full review right now, but

> We may want to combine this with cf-forms instead. Thoughts?

No, I don't think so. Things other than forms could trigger notifications.

src/cf-notification/src/cf-notification.less
cfarm: I'm noticing a pattern where we have the atomic prefix on the class names, but not on variables. I am fine with continuing this pattern.

src/cf-notification/src/cf-notification.less
cfarm: Don't include it! Use our brand colors instead. These are now available to any component importing cf-core - see #561

anselmbradford: Done!

anselmbradford:
Ready for final review!

anselmbradford:
> Loading icon and arrow look high. Should be vertically centered.

Unfortunately this is dependent on the icon used, so it will look good for some but not others. Flexbox could be investigated to correct this, but out of scope with this PR I think. I say we make it look best for the success, warning, and error icons and then investigate other layout techniques of the icons once the notification module is in.

> I recommend removing the exclamation points from examples (to stay within our editorial style)

Removed!

> Circle minicons are still slightly cut off on the left side.

This appears to be an antialiasing issue. I realized it will appear and disappear as I resize the page. @Scotchester any ideas on this one? I don't think there's anything we can do unless there's an anti-aliasing CSS property of some sort we could apply.

nataliafitzgerald:
@ans - Sounds good!

src/cf-notification/src/cf-notification.less
```diff
@@ -0,0 +1,44 @@
...
+   Notifications Styling
+   ========================================================================== */
+
+//
+// Theme variables
+//
+
+// Color variables
+
+@notification-bg:             #f7f8f9; // @gray-10
```
cfarm: these should use the variable as the value, not the hex code. up to you if you want to fix before merge, i'm fine with it coming later.

anselmbradford: Ah right, I can fix and then will merge.

cfarm:
:shipit: :shipit: :shipit: :shipit:

PR #574, opened 2017-08-30
Update package-lock

anselmbradford:
Updates package-lock files through `npm install` and `npm run cf-link`.


contolini:
Do lockfile updates trigger a version bump? That is, would this PR mean we have to increment the patch version of every component?

anselmbradford:
@contolini added changelog entries.

anselmbradford:
Deleted changelog

PR #578, opened 2017-09-01
Update web pack build flow

anselmbradford:
Updates build to use latest webpack.

## Additions

- Babel for ES6 transpiling.

## Removals

- `gulp-uglify` replaced with webpack uglify plugin.
- `gulp-webpack` replaced with `webpack`.

## Changes

- Convert build scripts to ES6

## Testing

- `npm run build && npm test` should pass.

## Note

- This removes the un-minified versions of `capital-framework.js` and `capital-framework.css`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
 ### Removed
 - **cf-tables:** [PATCH] Removed unused init method from `cf-table-row-links.js`.
+- **capital-framework:** [PATCH] Removed un-minified generated JS/CSS.
+- **cf-expandables:** [PATCH] Removed unused `bind`, `config`, and `group` vars.
 
 
 ## 4.8.5 - 2017-07-24
--- a/config/environment.js
+++ b/config/environment.js
@@ -1,0 +1,46 @@
+/* ==========================================================================
+   Settings for project environment.
+   Used by JavaScript gulp build process and JavaScript test configuration.
+   ========================================================================== */
+
+'use strict';
+
+/**
+ * @description
+ * Browser list for autoprefixer, see https://github.com/ai/browserslist.
+ * Support output lists can be viewed via copy/pasting into http://browserl.ist.
+ * @param {string} codeType - The code to support. Either 'js' or 'css'.
+ * @returns {array} List of supported browser strings.
+ */
+function getSupportedBrowserList( codeType ) {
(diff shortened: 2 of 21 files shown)


cfarm:
Do we want to keep the unminified CSS files as part of the released component? Is there a real use case for having this editable file? Are colleagues downloading this to use in third party software, editing it, etc?

anselmbradford:
Note this uglify warning is coming from AtomicComponent, and is fixed there in the latest release:

```
WARNING in cf-expandables.js from UglifyJs
Dropping unused variable timer [cf-expandables.js:510,6]
Dropping unused variable isAnimating [cf-expandables.js:512,6]
```

cfarm:
Works as advertised 👍 

@Scotchester @contolini was there a particular reason CF generated unminified CSS before? if not then this is good to merge.

Scotchester:
Unminified output can be very helpful for debugging purposes. I would recommend keeping it.

anselmbradford:
> Unminified output can be very helpful for debugging purposes. I would recommend keeping it.

Just comment out the minifier if this is needed, or we can add a flag to turn if off if this became a need.

Scotchester:
What is the benefit of not outputting it all the time?

anselmbradford:
> What is the benefit of not outputting it all the time?

There's a build complexity/performance benefit. The way it worked before was to bundle the file, rename it, save it, uglify it, rename it, save it again. With the updated webpack, uglifier is a plugin of webpack so happens at the bundling stage, so superficially the way around that I see to generate a minified and un-minified version would be to run webpack twice, which doesn't seem worth it considering how easy it is to just comment out uglifier and re-run `npm run build`. Also, I don't think this would be the correct approach to having a minified and un-minified version anyway. I think the correct way would be to have a production and development build process (set through a flag or otherwise) that could be switched between to generate production and dev assets.

Scotchester:
That makes sense. I'd like to see us institute the two types of processes, then, because I definitely run into periodic issues where debugging CSS is a pain because we don't output unminified versions, and sourcemaps still seem to be kinda fiddly. (Typically with cfgov-refresh.)

anselmbradford:
@Scotchester In this PR or another one?

Scotchester:
It should probably considered out-of-scope for this PR.

anselmbradford:
> It should probably considered out-of-scope for this PR.

I added a TODO. This also needs an update to add a webpack-config file, so that would be a good PR for that feature.

PR #579, opened 2017-09-05
Use CFPB brand colors in cf-pagination component

cfarm:
Adds CFPB brand colors as the default variable values for colors in `cf-pagination`. 

## Changes

- fix wrong border color for disabled state - was blue before, is now transparent (see screenshots below)
- fixed colors to be consistent with our brand palette and other components: uses same gray-text-on-gray10-background as form components, which is accessible. previous color implementation was not accessible here. the bg color used on cfgov-refresh is not a brand color. See example here: https://www.consumerfinance.gov/about-us/blog/

## Testing

In this repo:
- check out this branch
- run `npm run cf-link`

In a second local clone of this repo in a different folder:

- pull down and check out the `gh-pages-rm-overrides` branch: `git fetch && git checkout gh-pages-rm-overrides`
- if you haven't already in this second clone, [run all the install steps for the project](https://github.com/cfpb/capital-framework/tree/gh-pages#installation)
- run `npm run cf-link`
- run `npm run build`
- run `npm start`
- Compare colors on local site: 
   - http://localhost:3000/components/cf-pagination/ 
   - to the live cf documentation to make sure they match: https://cfpb.github.io/capital-framework/components/cf-pagination/

## Review

- @user

### Before - Capital Framework
screen shot 2017-09-05 at 3 33 30 pm

### Before - cfgov-refresh
screen shot 2017-09-05 at 3 33 50 pm

### After - Capital Framework
screen shot 2017-09-05 at 3 33 40 pm

## Notes

-

## Todos

- Markup for the newer/older buttons is different in capital framework docs than in cfgov-refresh (the pipe is missing from CF). I will make these consistent in a separate PR.

## Checklist

### DIFF
--- a/src/cf-pagination/src/cf-pagination.less
+++ b/src/cf-pagination/src/cf-pagination.less
@@ -8,15 +8,13 @@
 //
 
 // Color variables
-// `$color-` variables referenced in comments are from 18F's U.S. Web Design Standards
-// https://github.com/18F/web-design-standards/blob/staging/src/stylesheets/core/_variables.scss)
 
-@pagination-text-color:       #323a45; // $color-gray-dark
-@pagination-bg-color:         #f1f1f1; // $color-gray-lightest
+@pagination-text:               @gray;
+@pagination-bg:                 @gray-10;
 
 // Sizing variables
 
-@pagination-btn-min-width-px: 130px;
+@pagination-btn-min-width-px:   130px;
 
 
 //
--- a/src/cf-pagination/src/molecules/pagination.less
+++ b/src/cf-pagination/src/molecules/pagination.less
@@ -8,8 +8,8 @@
     &_form {
         padding: unit( 5px / @base-font-size-px, em );
         border-radius: unit( 4px / @base-font-size-px, em );
-        background: @pagination-bg-color;
-        color: @pagination-text-color;
+        background: @pagination-bg;
+        color: @pagination-text;
         text-align: center;
     }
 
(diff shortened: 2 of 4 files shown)


src/cf-pagination/src/molecules/pagination.less
```diff
@@ -8,8 +8,8 @@
     &_form {
         padding: unit( 5px / @base-font-size-px, em );
         border-radius: unit( 4px / @base-font-size-px, em );
-        background: @pagination-bg-color;
-        color: @pagination-text-color;
```
anselmbradford: I vaguely recall a discussion around including `-color` since `@pagination-text` on its own doesn't convey what it's doing (e.g. it could just as easily be a placeholder for the font size). But whatever is the precedent around here should be what we go with.

src/cf-pagination/src/cf-pagination.less
```diff
@@ -8,15 +8,13 @@
 //
 
 // Color variables
-// `$color-` variables referenced in comments are from 18F's U.S. Web Design Standards
-// https://github.com/18F/web-design-standards/blob/staging/src/stylesheets/core/_variables.scss)
 
-@pagination-text-color:       #323a45; // $color-gray-dark
-@pagination-bg-color:         #f1f1f1; // $color-gray-lightest
+@pagination-text:               @gray;
+@pagination-bg:                 @gray-10;
```
anselmbradford: TRIVIAL: These can be indented back a few spaces to the left and then the `@pagination-btn-min-width-px:` doesn't need the extra spaces.

src/cf-pagination/src/molecules/pagination.less
cfarm: yep i agree but i checked other components and we don't specify with such suffixes generally. a change i would be into, though.

PR #583, opened 2017-09-06
Add cf-modals

anselmbradford:
Features:

Modal can be closed with:
 - [x] A delete minicon or label in the upper right corner
 - [x] The next action button or link
 - [x] Anywhere outside the modal window
 - [x] ESC key on a keyboard

Accessibility:
- [x] The close minicon needs to be accessible from a keyboard.
- [x] The enter key should be mapped to any submission forms in the modal (hit enter while in form input).
- [x] Provide separate focus and hover states for the close minicon and any “next” action buttons.
- [x] Modal windows should be marked with aria-hidden="true" and toggled to false when visible.
- [x] Modals should be given the role=dialog ARIA role.
- [x] Add role="alertdialog" to the modal window requiring user interaction.
- [x] Code the close minicon using a button element for more semantic markup.
- [x] The heading of the modal dialog should be marked as an h1.

## Additions

- Adds cf-modals 1.0.

## Testing

- Check out this branch.
- Run `npm run cf-link`
- Change to another clone of capital-framework.
- Run `git fetch && git checkout gh-pages-modals`
- Run `npm link cf-modals`
- Check http://localhost:3000/components/cf-modals/
- Edit the page source to add `o-modal__visible` to the modal you want to show and reload the page.
- Inspect the page and see the aria- attributes toggling when hiding the modal.

## Todos:

These are taken from http://cfpb.github.io/design-manual/page-components/modals.html

- The find function (ctrl+F) will not search information contained within a modal window. _I don't think this is possible?_
- Include offscreen instructions that describe the modal dialog and how to interact with it. Offscreen interaction instructions should use aria-labelledby attribute. _I didn't find any information about this on https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role_
- When the modal dialog is displayed, the main content of the page should be marked with aria-hidden="true" to prevent screen readers from interacting with it.
- The tab order inside the alert dialog must wrap. 

## Screenshots

Dialog modal:
screen shot 2017-09-12 at 1 21 43 am

Alert modal:
screen shot 2017-09-12 at 1 21 26 am

## Notes

Modals requires some new additions to how CF works in regard to JS. Modals are absolutely positioned on the page, so can't be inside a container that has anything other than `position: static` set, otherwise it will end up relative to the container. This very likely means that there needs to be two components for modals: the actual modal and a "ModalTrigger" molecule (likely a link or button), which live in different areas of the page's code. The problem is we don't have a mechanism in place to communicate between the JS of two components on the page, which would be needed for a ModalTrigger to tell a Modal's JS to add the `o-modal__visible` class. @sebworks suggested that a redux utility be added to AtomicComponent and that @contolini had done work on this in cfgov-refresh already. This PR only provides the code structure for Modals, and were it to be visible, the logic for it to be closed.


src/cf-modals/src/Modal.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Handle a click to close the modal.
+ * Applied to the modal close button, backdrop, and offset.
+ */
+function onModalCloseClick() {
+
+  // TODO: Fix line is temporary till AtomicComponent sets this.ui.base
+  //       to equal this.element.
+  this.ui.base = this.element;
```
sebworks: I would just move this to the initialize method, so it only gets called once when the component is instantiated.

src/cf-modals/src/Modal.js
```diff
@@ -0,0 +1,57 @@
...
+
+  if ( this.insideClicked === false) {
```
sebworks: You might be able to use `toggleClass` here.

sebworks:
I would add code to handle closing the model on escape.

anselmbradford:
> I would add code to handle closing the modal when the escape key is pressed.

Good idea!

sebworks:
What is reasoning behind having the offset element?

jimmynotjim:
> Inspect the page and add o-modal__visible to the o-modal element to show the modal.

Can we add the trigger to the docs page before this gets published?

jimmynotjim:
Discussed in chat, but `position: fixed` should be used instead of `absolute`. A user shouldn't be able to scroll away from the modal without dismissing it, otherwise the page is covered by the gray overlay but it may be unclear why.

src/cf-modals/src/Modal.js
```diff
@@ -0,0 +1,57 @@
...
+const ModalOrganism = Organism.extend( {
+
+  insideClicked: false,
+
+  classes: {
+    isVisible: 'o-modal__visible'
+  },
+
+  events: {
+    'click .o-modal_close': 'onModalCloseClick',
```
jimmynotjim: Should values in `key: val` pairs within an object all be left aligned to match cfgov-refresh?

src/cf-modals/src/cf-modals.less
```diff
@@ -0,0 +1,91 @@
+/* ==========================================================================
+   Capital Framework
+   Modal Styling
+   ========================================================================== */
+
+//
+// Theme variables
+//
+
+// Color variables
```
jimmynotjim: You can remove this comment since we're no longer using unbranded colors

src/cf-modals/usage.md
```diff
@@ -0,0 +1,101 @@
...
+Theme variables for setting the color and sizes throughout the project.
+Overwrite them in your own project by duplicating the variable `@key: value`.
+
+### Color variables
+
+`$color-` variables referenced in comments are from 18F's
+[U.S. Web Design Standards](https://github.com/18F/web-design-standards/blob/staging/src/stylesheets/core/_variables.scss)
+
+```
+@modal-bg-outer:              #000000; // @black;
```
jimmynotjim: Update these to match the correct colors used above.

src/cf-modals/usage.md
```diff
@@ -0,0 +1,101 @@
...
+  - [Visible state](#visible-state)
+
+## Variables
+
+Theme variables for setting the color and sizes throughout the project.
+Overwrite them in your own project by duplicating the variable `@key: value`.
+
+### Color variables
+
+`$color-` variables referenced in comments are from 18F's
```
jimmynotjim: Remove this comment

src/cf-modals/usage.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Default state
+
+<div class="o-modal">
+    <div class="o-modal_backdrop"></div>
+    <div class="o-modal_container" aria-hidden="false" role="dialog">
+        <div class="o-modal_offset">
+            <div class="o-modal_content">
+                <div class="o-modal_inner">
+                    <p class="o-modal_close">
```
jimmynotjim: Is there a reason to wrap this in a paragraph? There's no content other than the button.

src/cf-modals/src/cf-modals.less
```diff
@@ -0,0 +1,91 @@
...
+   background-color: @modal-bg-inner;
+   padding: 30px;
+}
+
+.o-modal_actionbar {
+   background: @modal-action-bar;
+   padding: 10px 30px;
+}
+
+.o-modal_inner h1 {
```
jimmynotjim: I don't see an `h1` in the example. Is this necessary? Related, if it is, we should add a classname to the heading to keep with our Atomic principles.

src/cf-modals/usage.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+## Recommended modal pattern
+
+### Default state
+
+<div class="o-modal">
+    <div class="o-modal_backdrop"></div>
+    <div class="o-modal_container" aria-hidden="false" role="dialog">
+        <div class="o-modal_offset">
```
jimmynotjim: If the layout were updated to `position:fixed` I think we could drop this or the `_content` container.

src/cf-modals/usage.md
```diff
@@ -0,0 +1,101 @@
...
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
+```
+
+### Visible state
+
+Having the `o-modal__visible` modifier on the base element displays the modal.
```
jimmynotjim: Copy edit, suggest changing to "Adding the `o-modal__visible`..."

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 ## Unreleased
 
 ### Added
+- **cf-modals:** [MINOR] 1.0 added.
```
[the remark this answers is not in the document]
Scotchester: Based on what we learned from adding notifications, we should, I think, set the version in the `package.json` in this PR to `0.1.0` and set this to `[MAJOR]` so the script will bump it to 1.0.0.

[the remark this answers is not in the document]
contolini: Yeah I think that's the best bet but I would set it to `0.0.0` in package.json to emphasize that it hasn't been published yet.

anselmbradford:
> What is reasoning behind having the offset element?

I got rid of it!

> Can we add the trigger to the docs page before this gets published?

Needs to be another PR. We don't yet have the technology to communicate between two components.

anselmbradford:
@sebworks @jimmynotjim et al. This is ready for a review again. I made major adjustments based on some accessibility stuff I overlooked in the DM docs. See updated original post. Thanks!

sebworks:
I would say add the header as well, and move the close button into that.

src/cf-modals/usage.md
```diff
@@ -0,0 +1,175 @@
...
+A regular dialog modal does not contain a call to action for the user and
+can simply be closed without further interaction.
+
+<div class="o-modal"
+     aria-hidden="true"
+     role="dialog"
+     aria-labelledby="example-modal-1-title"
+     aria-describedby="example-modal-1-desc">
+    <div class="o-modal_backdrop"></div>
+    <div class="o-modal_container">
```
sebworks: Could you eliminate the container?

src/cf-modals/usage.md
```diff
@@ -0,0 +1,175 @@
...
+the `o-modal_close` class.
+
+<div class="o-modal"
+     aria-hidden="true"
+     role="alertdialog"
+     aria-labelledby="example-modal-2-title"
+     aria-describedby="example-modal-2-desc">
+    <div class="o-modal_backdrop"></div>
+    <div class="o-modal_container">
+        <form class="o-modal_content">
```
sebworks: should this be `o-modal_form`?

sebworks:
@anselmbradford, very nice work on this?

src/cf-modals/usage.md
anselmbradford: I didn't find a way. I just realized this essentially rebuilt the `o-modal_offset` that I removed earlier, lol. It's needed for properly handling the modal dimensions and the area that can be clicked to close the modal.

src/cf-modals/usage.md
anselmbradford: I don't think it needs to be. It only needs to support a form when there are inputs and a submit button, which may not always be the case. There's no extra styling outside of `o-modal_content`.

anselmbradford:
> I would say add the header as well, and move the close button into that.

It'll add an extra container, do we want that?

anselmbradford:
@jimmynotjim how's this look?

anselmbradford:
Can another dev review this? @Scotchester @contolini ? Jimmy's swamped right now.

cfarm:
Taking a look now.

cfarm:
Dialog modal - does the close button get focus by default when it's open, like with the input modal?

anselmbradford:
> Dialog modal - does the close button get focus by default when it's open, like with the input modal?

No, not according to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role

> Unlike regular alerts, an alert dialog must have at least one focusable control and focus must be moved to that control when the alert dialog appears. Generally alert dialogs have at least a Confirmation, Close or Cancel button that can be used to move focus to. Additionally, alert dialogs can have other interactive controls such as text fields, tabs or checkboxes. Which particular control focus should be moved to depends on the purpose of the dialog.

However, the way I coded it (hopefully) you could have an alertdialog without a form input, submit or button, and focus would move as a last resort to the close button.

cfarm:
That paragraph says that without those other inputs, focus should move to the close button by default. I think when they say "move to" they mean, we should code it so that's where the focus is when the dialog is open. So if this implementation isn't working that way, I would suggest making it so before we merge it. 

> Generally alert dialogs have at least a Confirmation, Close or Cancel button that can be used to move focus to.

cfarm:
When I add the `o-modal__visible` class to make these visible, this requirement is not happening:

- Modal windows should be marked with aria-hidden="true" and toggled to false when visible.

If there is another dependency needed before this can be the case, please update that in the PR body Todos section.

src/cf-modals/src/cf-modals.less
```diff
@@ -0,0 +1,89 @@
...
+
+//
+// Theme variables
+//
+
+// .o-modal
+@modal-backdrop:   @black;
+@modal-body:       @white;
+@modal-footer:     @gray-5;
+@modal-border-top: green;
```
cfarm: missing `@` variable indicator, should be `@green`

src/cf-modals/usage.md
```diff
@@ -0,0 +1,175 @@
...
+Theme variables for setting the color and sizes throughout the project.
+Overwrite them in your own project by duplicating the variable `@key: value`.
+
+### Color variables
+
+```
+@modal-backdrop:   @black;
+@modal-body:       @white;
+@modal-footer:     @gray-5;
+@modal-border-top: green;
```
cfarm: fix green variable here too

cfarm:
also, when I open the Alert modal, the input field does not have focus - still have to tab or click. Please clarify whether this functionality is to be expected with this implementation or if it's coming in the JS work.

cfarm:
Good news, these look great in ie8! Haven't tested other browsers yet. Did notice that the Capital Framework docs styles aren't working in ie8 for some reason though, probably an issue on all branches...

screen shot 2017-09-15 at 3 49 27 pm
screen shot 2017-09-15 at 3 50 30 pm

Scotchester:
If we switch the media query from `min-width` to `max-width`, we can get the appropriate width in IE8.

anselmbradford:
@cfarm 

> That paragraph says that without those other inputs, focus should move to the close button by default. I think when they say "move to" they mean, we should code it so that's where the focus is when the dialog is open.

It should be working this way, so if the other inputs are not present focus will be on the close button when it is open.

> When I add the o-modal__visible class to make these visible, this requirement is not happening:

Are you by chance adding this class via the inspector once the page is loaded? This may not be clear enough in the testing instructions, but you need to add it to the page source temporarily and reload the page. Since the focus and aria- stuff depends on JS the JS won't run unless `o-modal__visible` is present at load time or if the `show()` function is called. A separate modal trigger component would need to be programmed to call the `show()` method on its associated modal, but that's not part of the PR. If you are adding the class to the source and this is still borken, let me know!

src/cf-modals/src/cf-modals.less
anselmbradford: Fixed! Thx!

src/cf-modals/usage.md
anselmbradford: Fixed! Thx!

anselmbradford:
@Scotchester Swapped the media query.

jimmynotjim:
@anselmbradford if the trigger is outside this PR, can we make the modal visible by default in the doc site? If I'm looking for an example, I'd rather be able to see it on GH Pages than have to edit my local source. I'd also like to see a note in here about the trigger being it's own thing.

anselmbradford:
> @anselmbradford if the trigger is outside this PR, can we make the modal visible by default in the doc site? If I'm looking for an example, I'd rather be able to see it on GH Pages than have to edit my local source. I'd also like to see a note in here about the trigger being it's own thing.

@jimmynotjim It's not really possible, there are two modals that would overlap and the modals cover the whole page, so it would look hella broken. I know the testing here is a pain, but this PR is really just to show the markup only to sync with DM and set the groundwork for showing an example. That said I want the example to work, which is why the testing instructions are there, but until we flesh out how we want components to communicate with each other I'd omit that part of it completely. Projects like Mosaic can still use the markup and classes as reference since they have their own triggering mechanism.

cfarm:
OK I was adding it via the inspector. Trying it our properly, this is the behavior I see:

## Dialog model

Expected behavior:
- Focus will be on the close button when it is open.

Actual behavior:
- Focus is not on close button when the dialog is open. If I hit "Enter," nothing happens. I have to use the tab key to get focus on the close button. 
- aria-hidden works as expected

## Alert model

Focus works as expected, aria-hidden="false" works as expected

cfarm:
I think I understand why we can't add any triggers to this PR, but can we get a demo just to clarify?

anselmbradford:
@cfarm 

As I understand it from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role) the dialog modal does not steal focus, even for the close button, unlike the alert dialog:

> Unlike regular alerts, an alert dialog must have at least one focusable control

src/cf-modals/src/cf-modals.less
```diff
@@ -0,0 +1,91 @@
...
+   Capital Framework
+   Modal Styling
+   ========================================================================== */
+
+//
+// Theme variables
+//
+
+// .o-modal
+@modal-backdrop:   @black;
```
cfarm: should be gray-80 according to Lorelei's DM updates: https://ielerol.github.io/design-manual/page-components/modals.html#style

src/cf-modals/src/cf-modals.less
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    &_backdrop {
+        position: fixed;
+        width: 100%;
+        height: 100%;
+        top: 0;
+        left: 0;
+        background-color: @modal-backdrop;
+        opacity: 0.5;
```
cfarm: opacity should be set to 0.7 according to Lorelei's DM updates: https://ielerol.github.io/design-manual/page-components/modals.html#style

src/cf-modals/src/cf-modals.less
ielerol: Could we get screenshots comparing this difference? I discussed with Natalia and we're not sure which version is correct. @anselmbradford do you know any history for why the modals code is different than the spec?

anselmbradford: @ielerol I don't have any further history on this part, but here's a comparison of before/after. Let me know which to keep.

As is:
 screen shot 2017-09-12 at 1 21 26 am

Lighter backdrop:
screen shot 2017-09-28 at 4 40 47 pm

ielerol: Thanks @anselmbradford! We thinker the lighter example is working better, so let's go with the gray-80 and .7 opacity

anselmbradford: Done!

src/cf-modals/src/cf-modals.less
anselmbradford: Fixed in https://github.com/cfpb/capital-framework/pull/583/commits/d1674b14e2f9f7262c8cbbef6fa13001bf2de7a6

jimmynotjim:
@anselmbradford sorry to disappear, where are we on this? It looks like there was still question about focus, but maybe that's resolved cc @cfarm? Are we planning on merging without tests? If so, lets make sure to add an issue to get those in soonish.

anselmbradford:
> sorry to disappear, where are we on this? It looks like there was still question about focus, but maybe that's resolved cc @cfarm? Are we planning on merging without tests? If so, lets make sure to add an issue to get those in soonish.

Planning to merge into cfv5 prerelease branch today and create a roadmap of issues for that release.

anselmbradford:
Okay created https://github.com/cfpb/capital-framework/projects for v5. I'll merge this into `prerelease-ver-5` branch if there are no objections. @cfarm did I address your code review?

PR #592, opened 2017-09-07
Breaks badges into two rows

anselmbradford:
## Changes

- CF modules have become too numerous for the npm badges to fit on one row. This breaks them into two rows.


jimmynotjim:
@anselmbradford Unrelated, but any idea why cf-notifications is showing as 0.0.1 instead of 1.0.1?

PR #593, opened 2017-09-07
Rename cf-notification to cf-notifications

anselmbradford:
## Changes

- Rename cf-notification to cf-notifications


jimmynotjim:
@contolini Will this break npm or the release script?

CHANGELOG.md
```diff
@@ -4,13 +4,14 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 ## Unreleased
 
 ### Added
-- 
+-
 
 ### Changed
-- 
+- **cf-notifications:** [PATCH] Renamed from `cf-notification`
+to `cf-notifications`.
```
contolini: Do we enforce 80 character line lengths in markdown files?

src/cf-notifications/package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "cf-notification",
-  "version": "1.0.1",
-  "description": "Capital Framework notification",
-  "less": "src/cf-notification.less",
-  "style": "cf-notification.css",
+  "name": "cf-notifications",
+  "version": "1.0.2",
```
contolini: Don't manually increment the version. When the bot opens the release PR it'll do it for you based on the changelog entry you added above.

CHANGELOG.md
anselmbradford: I usually wrap them. The only practical reason to enforce them at all is to prevent scrolling in the GH UI, right, which applies to all text files?

src/cf-notifications/package.json
anselmbradford: Fixed!

contolini:
> @contolini Will this break npm or the release script?

Hopefully not. 😰

CHANGELOG.md
contolini: I thought it had to do with diffing and that markdown files should be have newlines at the end of discrete thoughts.

anselmbradford: Should I revert it? I've been wrapping lines in cfgov-refresh changelog for ages 😱

Scotchester: As you both noted, there are two aspects to this:
1. Breaking lines at discrete thoughts can make diffing future changes easier.
2. Some max length should be adhered to so that lines aren't horizontally scrolling in the GH UI.

The two are mostly unrelated. The max length could be longer than 80 characters, but I don't mind using that as our guideline, since that's what we enforce for CSS, JS, and Python.

In this case, I would ask that successive lines of bullets be indented to match the first non-bullet-non-whitespace character of the first line, though. Makes reading lists in raw Markdown easier.

contolini: @anselmbradford It doesn't matter to me. Merge at will.

PR #594, opened 2017-09-07
Rename cf-notification to cf-notifications in docs

anselmbradford:
## Changes

- Rename cf-notification to cf-notifications in docs.


src/static/css/main.less
```diff
@@ -13,7 +13,7 @@
 @import (less) "cf-layout.less";
 @import (less) "cf-buttons.less";
 @import (less) "cf-forms.less";
-@import (less) "cf-notification.less";
+@import (less) "cf-notifications.less";
```
Scotchester: Won't this require updating the dependency in `package.json` to be `cf-notifications`?

anselmbradford: Uh yup, fixed!

PR #601, opened 2017-09-08
Adds state sidecar .hover to basic inputs

anselmbradford:
`a-btn__link` have a `hover` class for manually triggering the hover state. This adds that to basic form inputs.

## Additions

- Adds sidebar `.hover` class to support manually triggering the hover/focus state in documentation.

## Testing

- `git fetch && git checkout form_states`
- `npm run cf-link`
- Switch to a CF clone repo.
- `git fetch && git checkout gh-pages-form_states`
- `npm run cf-link`
- `npm run build && npm start`

## Screenshots

screen shot 2017-09-08 at 3 16 55 pm

screen shot 2017-09-08 at 3 17 01 pm

## Notes

- I converted
```
&:hover:before {
        border-color: @input-border__focused;
}
```
…to…
```
&:hover:before,
&.hover:before {
        border-color: @input-border__hover;
}
```
…which made me realize we have variables for `@input-border__focused` and `@input-border__hover`, but since the CSS for those two states is separate in the basic inputs those variables will not work as expected in regard to checkboxes and radio buttons. Opened issue https://github.com/cfpb/capital-framework/issues/600

PR #602, opened 2017-09-08
Adds basic input states documentation

anselmbradford:
Documentation for https://github.com/cfpb/capital-framework/pull/601

## Notes

- @jimmynotjim So if this includes my local linked CF, does it matter when a release goes out, since my only changes are the changes in https://github.com/cfpb/capital-framework/pull/601 ?


jimmynotjim:
@anselmbradford normally I don't think we should be releasing doc site changes for unreleased code, but in this case, it's not a big deal since the changes are just for documentation.

My biggest worry with having symlinked components in your doc site is accidentally falling behind (like say working from a stale branch) or unintentionally including unreleased code, especially if the unreleased changes were to stall during review.

PR #603, opened 2017-09-11
Updates qunit to 1.0.0 from 0.7.7

anselmbradford:
## Changes

- Updates qunit to 1.0.0 from 0.7.7

## Testing

- `npm install && npm test` should pass.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "capital-framework",
-  "version": "4.9.1",
+  "version": "4.9.2",
   "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


jimmynotjim:
Since we aren't using jQuery anymore, should we be moving away from qUnit for testing?

anselmbradford:
> Since we aren't using jQuery anymore, should we be moving away from qUnit for testing?

Yes I think so. It would be easier if the testing framework was consistent between here, cfgov-refresh, and DM. Qunit was still out of date for the time being though.

PR #615, opened 2017-09-20
a-text-input updates

anselmbradford:
## Additions

- Add color variable for text input.
- Add sidecar `.hover` classes to text input.

## Removals

- Convert `a-text-input__focus` to `.focus` class.

## Changes

- Lighten disabled input gray.
- Adds markdown code block language to docs for syntax highlighting (we should do this across the board for all modules).
- Adds placeholder text to inputs in docs.

## Testing

## Screenshots

2017-09-20-15-11-localhost-3000

## Notes

- Talked with @jimmynotjim about having a build environment for documentation, which in other environments `.hover` and `.focus` would be stripped out.
- Do we need `a-text-input__disabled`?


anselmbradford:
Added an update to cancel hover in the disabled state and to set the text to gray in that state, per DM specs.

screen shot 2017-09-20 at 3 53 41 pm

jimmynotjim:
Looking around I'm pretty sure I was wrong when I thought we used `a-text-input__disabled`. I vote we move this to a chained class specifically for docs only as well.

anselmbradford:
> Looking around I'm pretty sure I was wrong when I thought we used a-text-input__disabled. I vote we move this to a chained class specifically for docs only as well.

We don't even need a chained class, right? We can just use the `disabled` attribute. Hover and focus need a chained class because there isn't an attribute we can use.

nataliafitzgerald:
@jimmynotjim - YES. We definitely want to break this up if we can get some resources to work on this. This page is insanely long.

PR #616, opened 2017-09-20
Separate Table class from initialization code

anselmbradford:
Table class was wrapped up with the initialization code. This separates it like is done with Expandables.

## Changes

- Separate Table class from initialization code.
- Rename class modules TableRowLinks and TableSortable to match their export type.

## Testing

- `npm run build` should pass.

## Notes

- I'll ES6-ify it in another PR so the changes are easier to see.


sebworks:
@anselmbradford, we shouldn't call `init` at all. I don't think we need `Table.js`.

anselmbradford:
> we shouldn't call init at all. I don't think we need Table.js.

You mean `cf-tables.js`, right? And yeah you're right, but let's strip it out all at once in another PR. Expandables needs it gone too, and cfgov-refresh would need an update that corresponds with the PR, which we don't want for this release.

PR #627, opened 2017-09-29
Change font size in helper text to use variable.

anselmbradford:
## Changes

- Change font size in helper text to use variable.

## Testing

## Screenshots

screen shot 2017-09-29 at 11 32 28 am

### DIFF
--- a/src/cf-forms/src/atoms/label.less
+++ b/src/cf-forms/src/atoms/label.less
@@ -4,7 +4,7 @@
     &_helper {
         .u-webfont-regular();
         color: @label-helper;
-        font-size: unit( 16px / @base-font-size-px, em );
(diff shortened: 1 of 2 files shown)


src/cf-forms/src/atoms/label.less
```diff
@@ -14,5 +14,9 @@
 
         // Overwrites heading-4 margin.
         margin-bottom: unit( 10px / @font-size, em );
+
+        &_helper {
```
jimmynotjim: This will need it's full class name otherwise it'll resolve to `a-label__heading_helper` rather than `a-label__heading a-label_helper` that you want

anselmbradford: Oh shoot! Dumb mistake. Fixed!

PR #630, opened 2017-10-03
Fixes docs to include single wrappers for full-width text inputs

anselmbradford:
## Changes

- Docs: Nests full width modifier inputs in their own m-form-field divs.

### DIFF
--- a/src/cf-forms/usage.md
+++ b/src/cf-forms/usage.md
@@ -314,7 +314,9 @@ Generally this is only useful for documentation purposes.
            type="text"
            id="full-textinput-example"
            value="Lorem ipsum">
-    <br><br>
+</div>
+<br>
+<div class="m-form-field">
     <label class="a-label a-label__heading" for="full-textarea-example">
         A full-width textarea input
     </label>
@@ -331,7 +333,9 @@ Generally this is only useful for documentation purposes.
            type="text"
            id="full-textinput-example"
            value="Lorem ipsum">
-    <br><br>
+</div>
+<br>
+<div class="m-form-field">
     <label class="a-label a-label__heading" for="full-textarea-example">
         A full-width textarea input
     </label>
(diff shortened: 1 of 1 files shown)


src/cf-forms/usage.md
```diff
@@ -331,7 +333,9 @@ Generally this is only useful for documentation purposes.
            type="text"
            id="full-textinput-example"
            value="Lorem ipsum">
-    <br><br>
+</div>
+<br>
```
Scotchester: For the code example, should we omit the `<br>` and just have a blank line here?

jimmynotjim: It's from way back before these were ever wrapped. I'd say yes

anselmbradford: Does not work:
screen shot 2017-10-03 at 6 00 44 pm

Scotchester:
Maybe I commented in the wrong place, but I'll taking about the sample
code, not the coffee that renders the live example.

anselmbradford:
> Maybe I commented in the wrong place, but I'll taking about the sample
code, not the coffee that renders the live example.

Ooo. Do we want to go down the road of where the code example and rendered code are different? I'm not sooo keen on that. I'd think it shouldn't be confusing of why copying the code example does not look exactly like the live example. If we don't want the `<br>` what do you think about providing a code example under each instead of one code example combined together?

Scotchester:
I think realistically, people won't be copying both fields together, they'll select one or the other and copy that. But if we want to be 0% error-prone, I agree with the idea of doing the code example under each.

anselmbradford:
> I think realistically, people won't be copying both fields together, they'll select one or the other and copy that. But if we want to be 0% error-prone, I agree with the idea of doing the code example under each.

Done!

anselmbradford:
screen shot 2017-10-04 at 11 02 00 am

Scotchester:
:shipit:

PR #631, opened 2017-10-03
Multiline form fields

anselmbradford:
Uses CSS grid to add support for multi-line labels and helper text in checkboxes/radio buttons. Wraps in unsupported browsers.

## Changes

- Add multi-line support to checkboxes/radio buttons.

## Testing

## Screenshots

See "Inputs helper text" section:
2017-10-03-17-51-localhost-3000

## Note
We may want to merge this into CFv5 branch, as all the supported environments could take some time to test, and could present browser-specific bugs as this has a lot of nuance.


anselmbradford:
> Still not a fan of giant paragraphs of label text but I have seen where it's broken onto two lines and makes sense so lets go with it.

They could be shortened. I just wanted ensure they wrap to multiple lines for the example.

> As far as cfv5, what if you used flexbox instead of grid? We're already supporting flexbox in a few places via @supports, we could probably do that here too.

Yeah the original PR for this was flexbox-based and I tried again with this update, the problem is the helper text. We need a layout like:

------------
|      |     b    |
|  a  |-------
|      |     c    |
------------ 

a = icon
b = label
c = helper text

The markup structure is:
<label>
  ::before (a)
  (b) 
  <small> (c)

Since flexbox is essentially wrapped rows or columns of content, I couldn't find a way to get (c) below (b) without adjustments to the markup. With grid you can place the content where ever you want within the grid.

nataliafitzgerald:
@anselmbradford 
For the CF documentation example, do we need to _show_ multiple lines? I think it's good that we're addressing the indentation issue for multiple lines but shouldn't we show the most common use case (knowing that should a label wrap to two lines, we're covered in terms of the styling)? Showing multiple lines in the documentation seems to suggest that we support having really, really long labels like this. Or are you just showing it this way so that we have screenshots for the purposes of review? If that's the case then you can ignore the last few sentences here. ;)

anselmbradford:
> For the CF documentation example, do we need to show multiple lines? I think it's good that we're addressing the indentation issue for multiple lines but shouldn't we show the most common use case (knowing that should a label wrap to two lines, we're covered in terms of the styling)? Showing multiple lines in the documentation seems to suggest that we support having really, really long labels like this. Or are you just showing it this way so that we have screenshots for the purposes of review? If that's the case then you can ignore the last few sentences here. ;)

I wanted to ensure it spanned multiple lines, but I can shorten the label filler text. I think we do want an example in the docs since it does appear in the wild (in Mosaic) and when changes are made in the future we can easier see if something was broken.

Scotchester:
Can we add a `max-width` rule to keep multi-line label line lengths reasonable?

nataliafitzgerald:
@Scotchester  - If your proposed change gets us closer to vertical centering of that checkmark in Chrome and Safari I'm in full support.

anselmbradford:
> Can we add a max-width rule to keep multi-line label line lengths reasonable?

@Scotchester Added https://github.com/cfpb/capital-framework/pull/637, does that work?

anselmbradford:
Shortened text in examples:
screen shot 2017-10-05 at 11 49 25 am

nataliafitzgerald:
@anselmbradford 
Once the max width goes into effect we'll need to cut down on the number of characters here for the label, otherwise it will go to more than 2 lines.

anselmbradford:
> Once the max width goes into effect we'll need to cut down on the number of characters here for the label, otherwise it will go to more than 2 lines.

Okay shortened them further, though it's going to depend a bit on the window's size as to whether they're 2 or more lines.

nataliafitzgerald:
@anselmbradford 
If we can just get it to 2 lines at the max width we should be fine.

scripts/gulp/styles.js
```diff
@@ -21,7 +21,7 @@ function stylesCf() {
     .pipe( gulpAutoprefixer( {
       browsers: BROWSER_LIST.LAST_2_PLUS_IE_8_AND_UP
     } ) )
-    .pipe(gulpCssmin())
+    //.pipe(gulpCssmin())
```
Scotchester: Need to un-comment this.

src/cf-forms/src/molecules/form-fields.less
```diff
@@ -11,11 +11,19 @@
     &__checkbox,
     &__radio {
         .a-label {
+            display: -ms-inline-grid;
+            display: inline-grid;
+            // 30px is width of checkbox/radio button plus the needed padding.
+            -ms-grid-columns: unit( 30px / @base-font-size-px, em ) 1fr;
+            grid-template-columns: unit( 30px / @base-font-size-px, em ) auto;
```
Scotchester: Will Autoprefixer not handle the `-ms-` prefixes?

jimmynotjim: It should.

nataliafitzgerald:
@anselmbradford 
Can you update the screenshot to show the max-width (670px)?

src/cf-forms/src/molecules/form-fields.less
anselmbradford: Notice the values `1fr` and `auto` are different. IE has been what's holding up this PR. It absolutely mangles the grid. I've decided to just drop the columns entirely in IE and go with a fallback state where the label doesn't have the indent.

anselmbradford:
Okay, this is ready for review again. It was held up due to a difference in IE that I decided to provide a fallback state for, which I hope is acceptable. The IE state wraps the text, but doesn't indent it and the helper text falls on the same line, if present.

Note also that the grid shuffles the layout slightly, so there is a 1px shift in the positioning of checkboxes and radio buttons. See the screenshots below of production docs, this PR's docs, and an overlay of the difference that shows the shift. You'll want to open the diff image and view full-size and scroll to the middle-ish of the image to see the checkboxes and such. Everything at Form Validation and below is just showing up in the diff is just because the whole content has been pushed down.

| Before  | After   | Difference   |
|---|---|---|
|  2017-12-22-12-53-cfpb github io | 2017-12-22-12-51-localhost_3000 | 2017-12-22-12-51-localhost_3000-diff  | 

Latest Edge 16 looks like this:
screen shot 2017-12-22 at 1 27 12 pm

Older Edge, IE11 and lower look like this:
screen shot 2017-12-22 at 1 29 08 pm

Firefox:
screen shot 2017-12-22 at 1 31 30 pm

Chrome 55 for Android (should be supported in Chrome 62):
screen shot 2017-12-22 at 2 00 29 pm

iOS 11.1 (iPhone X):
screen shot 2017-12-22 at 2 08 28 pm

iOS 8.1 (iPad 2):
screen shot 2017-12-22 at 2 16 21 pm

iOS 10.3 (iPhone 7):
screen shot 2017-12-22 at 2 28 44 pm

Safari:
screen shot 2017-12-22 at 2 09 14 pm

anselmbradford:
Updated helper text to include parentheses per request from @nataliafitzgerald 
screen shot 2017-12-22 at 4 18 39 pm

anselmbradford:
🕺

PR #632, opened 2017-10-04
Block helper txt

anselmbradford:
## Additions

- Add `a-label_helper__block` modifier to create block helper text below a label.

## Testing

## Screenshots

screen shot 2017-10-04 at 11 50 34 am

### DIFF
--- a/src/cf-forms/src/atoms/label.less
+++ b/src/cf-forms/src/atoms/label.less
@@ -8,11 +8,10 @@
 
         &__block {
             display: block;
-        }
-    }
 
-    & > &_helper__block {
-        margin-top: unit( 10px / @base-font-size-px, em );
+            // Add a gap between the label helper and label.
+            margin-top: unit( 10px / @base-font-size-px, em );
+        }
     }
 
     &__heading {
--- a/src/cf-forms/src/atoms/label.less
+++ b/src/cf-forms/src/atoms/label.less
@@ -5,6 +5,14 @@
         .u-webfont-regular();
         color: @label-helper;
         font-size: unit( @size-v / @base-font-size-px, em );
+
+        &__block {
+            display: block;
+        }
+    }
+
+    & > &_helper__block {
+        margin-top: unit( 10px / @base-font-size-px, em );
     }
 
     &__heading {
--- a/src/cf-forms/usage.md
+++ b/src/cf-forms/usage.md
@@ -133,15 +133,34 @@ Color variables referenced in comments are from [cf-core cf-brand-colors.less](h
 
 ### Label helper text
 
+#### Inline helper text
+
+Appears with label headings.
 Used for designating an input as optional for user input.
 
 <label class="a-label a-label__heading">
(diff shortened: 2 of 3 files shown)


src/cf-forms/src/atoms/label.less
```diff
@@ -5,6 +5,14 @@
...
         color: @label-helper;
         font-size: unit( @size-v / @base-font-size-px, em );
+
+        &__block {
+            display: block;
+        }
+    }
+
+    & > &_helper__block {
+        margin-top: unit( 10px / @base-font-size-px, em );
```
jimmynotjim: Why not put this in with the display rule?

anselmbradford: I was ensuring that the margin is *only* applied when the helper is inside a label. 
It should always be so I guess they can be combined.

anselmbradford: Fixed in https://github.com/cfpb/capital-framework/pull/632/commits/147df19d09ea11c0d98c83d8f1d9488908510bb8

jimmynotjim: Yeah, since it's an `_element` it should never be used outside the `a-label` parent

src/cf-forms/usage.md
```diff
@@ -133,15 +133,34 @@ Color variables referenced in comments are from [cf-core cf-brand-colors.less](h
 
 ### Label helper text
 
+#### Inline helper text
+
+Appears with label headings.
```
Scotchester: Does this mean inline helpers are not to be used with standard labels?

jimmynotjim: cc @nataliafitzgerald I believe you discussed this with Ans because I brought it up when he was first exploring options a couple weeks ago.

Scotchester: 'cause if so, what would the pattern be for indicating that a field with a basic label is required?

jimmynotjim: From what I remember in the conversation, we don't use standard labels outside of the ones to the side of a checkbox/radio. Everything else is a label heading or a legend, which makes me realize we probably need this for legends as well...

anselmbradford: What Jimmy said. Helper text appears in large target checkboxes/radio buttons, but those use block helpers.

Scotchester: Alright, I don't want this discussion to take this PR off the rails, but it sounds like `a-label__heading` should be the standard `a-label` and we should have a modifier for the one that sits next to a check/radio, like `a-label__inline` or something.

Scotchester: (Just now realized it was merged before I commented the first time.)

anselmbradford: You bring up a good point @Scotchester, the general standard is `label__heading`, so why should we need to add that all the time.

jimmynotjim: 🤔 wicked smaht

Issue #207, opened 2015-11-04
[cf-core] Max-Widths on paragraph


KimberlyMunoz:
There was discussion on setting a rule on max-width with ems for long headings and/or paragraphs. The use case is that it would prevent paragraphs from getting past a readable length.

https://github.com/cfpb/cf-core/issues/25

jimmynotjim:
This was [addressed by the Platform team](https://github.com/cfpb/cfgov-refresh/pull/2475/files#diff-b6e78e12f3348c69634c6b95cd89390cR210) and should be migrated to this project.

PR #637, opened 2017-10-05
Set max-width for top-level elements

anselmbradford:
Fixes https://github.com/cfpb/capital-framework/issues/207 and https://[GHE]/CFGOV/platform/issues/727

## Changes

- Set max-width on top-level elements.

## Testing

- Check any paragraph, dd, dt, heading, list, or label and edit the source in the inspect to add long text as needed then and/remove this style rule to see before/after.

## Screenshot

 Before:
screen shot 2017-10-05 at 11 36 45 am

After:
screen shot 2017-10-05 at 11 36 50 am

## Notes

- This differs from what's in cfgov-refresh by including labels.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 -
 
 ### Changed
--
+- **cf-layout:** [MINOR] Set max-width on top-level elements.
 
 ### Removed
 -
--- a/src/cf-layout/src/cf-layout.less
+++ b/src/cf-layout/src/cf-layout.less
@@ -232,6 +232,21 @@
     &:extend( .wrapper all );
 }
 
+.content_main,
+.content_intro {
+    dd,
+    dt,
+    h3,
+    h4,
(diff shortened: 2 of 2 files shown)


anselmbradford:
Hmmm… applying this to the label truncates the large target labels too. Not sure if this is a problem or not. Maybe we do want them narrower?

Before:
screen shot 2017-10-06 at 10 04 48 am

After:
screen shot 2017-10-06 at 10 03 37 am

anselmbradford:
@nataliafitzgerald said the truncated width in labels is fine (correct me if I'm wrong).

anselmbradford:
> Nice work. Can you make sure to delete this from cfgov-refresh when it's updated to use whatever release this change ends up in?

You bet!

ielerol:
I agree with Natalia, I think the label truncation is fine, and I'd prefer that the label box match the text width, generally.

anselmbradford:
> Nice work. Can you make sure to delete this from cfgov-refresh when it's updated to use whatever release this change ends up in?

👍  I'll do a release after this.

PR #639, opened 2017-10-05
Remove inline code margin

anselmbradford:
[This PR](https://github.com/cfpb/capital-framework/commit/94744fdc7bae8ffaa90d6d63a545d2aab95d299d) added this margin, but it messes with the line height of paragraphs across the board that have inline code blocks. I couldn't find where it was originally used. @Scotchester do you remember? Can it be removed?

## Removals

- Removes margin above inline code blocks.

## Testing

- `npm run build && npm start` and view the component pages vs the live site.

## Screenshots

Before:
screen shot 2017-10-05 at 3 18 01 pm

After:
screen shot 2017-10-05 at 3 18 09 pm

### DIFF
--- a/src/static/css/main.less
+++ b/src/static/css/main.less
@@ -54,10 +54,6 @@ nav ul {
     padding: 0;
 }
 
-.highlighter-rouge {
-    margin-top: unit( 15px / @base-font-size-px, em );
-}
-
 // custom modules
 @import (less) "header.less";
 @import (less) "hero.less";


Scotchester:
I believe the intent was to make sure that full block code examples had some margin above them, and I probably just didn't consider/notice that it would apply to inline code, as well. Is there a way to scope it to blocks of code only?

anselmbradford:
We may not need to remove it now that #640 is in. I'll release and update the docs and we can check.

anselmbradford:
Docs updated, yeah doesn't look like we need to remove this at this time.

Issue #555, opened 2017-08-09
[Bug] Inline `code` pieces are having excess bottom margin applied.


marteki:
As seen on pages for the Design Manual, `code` tags that are placed inline with paragraph text are having extra bottom margin applied where it shouldn't be.

## Current behavior

- The code tags appear to be superscript compared to the other text in the sentences.

## Expected behavior

- The code tags run inline with the other text in the sentences.

## Steps to replicate behavior

1. Go to http://cfpb.github.io/design-manual/best-practices/accessibility-best-practices.html#perceivable
2. Look at the paragraph copy in this and other sections on the page.

## Screenshots

__Current__

![screen shot 2017-08-09 at 3 25 42 pm](https://user-images.githubusercontent.com/1183435/29141228-4d8a1d16-7d1b-11e7-8c1b-2deb0fbe940c.png)

__Expected__

For help in troubleshooting and determining what caused this:
The margin disappears if the `overflow-x` and `margin-bottom` attributes are turned off in the `.content code` CSS, along with the `margin-bottom` attribute in the general `code` CSS section.

![screen shot 2017-08-09 at 3 34 56 pm](https://user-images.githubusercontent.com/1183435/29141238-534296f2-7d1b-11e7-8d6b-e4b365505241.png)

Scotchester:
It's being caused by an unfortunate combination of [this line in CF](https://github.com/cfpb/capital-framework/blob/canary/src/cf-core/src/cf-base.less#L554) and [this line in the DM's CSS](https://github.com/cfpb/design-manual/blob/13a279827ce45b5f1a5fc07d7670fbf721f7d9a4/src/static/css/content.less#L123).

As for the CF code: @jimmynotjim Do you know why you needed the `display: inline-block;` for `code` elements? A cursory test seems to indicate that it's not necessary.

Separately, that `margin-bottom` in the DM styles should likely be on a whole `pre` block, not on any ol' `code` element.

jimmynotjim:
I do not remember why `inline-block` was needed or if it even was. I feel like I took most of that code from another example, possibly GitHub itself, but it doesn't seem to be a style rule they're using, so lets drop it.

marteki:
Was this fixed? I still see it happening on http://cfpb.github.io/design-manual/page-components/tables.html

![screen shot 2017-09-25 at 3 45 10 pm](https://user-images.githubusercontent.com/1183435/30827780-3cbaa70e-a209-11e7-8bee-a732754e28b8.png)

PR #640, opened 2017-10-05
Remove inline block style from inline code

anselmbradford:
Fixes https://github.com/cfpb/capital-framework/issues/555

## Removals

- Removes `display: inline-block` from inline `code` elements.

## Testing

- Look at inline code on the docs pages and compare to live site.

## Note
- Inline-block added a gap, so maybe https://github.com/cfpb/capital-framework/pull/639 isn't necessary with this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 -
 
 ### Changed
-- **cf-core:** [MINOR] Remove inline-block style from inline code.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 -
 
 ### Changed
--
+- **cf-core:** [MINOR] Remove inline-block style from inline code.
```
Scotchester: Actually, this should probably be a patch.

anselmbradford: Fixed!

PR #651, opened 2017-10-17
Adds TOC to cf-grid

anselmbradford:
cf-grid was the only module missing a TOC in its usage docs.

## Additions

- Adds TOC to cf-grid.


anselmbradford:
> It's 100% fine with me if you want this to be a silent docs update that rolls out with the next release

I'm fine with that!

PR #659, opened 2017-10-18
Change example links to alerts

anselmbradford:
## Changes

- Change links in table row example to an alert instead of an actual URL link so that clicks of the example don't leave the docs page.
- Make example code block actually follow implemented code for row links in tables.

## Testing

- Set up local test environment and check docs for cf-tables and click a row link https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-components-locally

## Screenshot

screen shot 2017-10-18 at 11 50 22 am

### DIFF
--- a/src/cf-tables/usage.md
+++ b/src/cf-tables/usage.md
@@ -189,14 +189,14 @@ highlighting and hyperlinking rows which contain links.
     <tbody>
         <tr>
             <td data-label="Column 1">
-                <a href="https://example.com/">Example 1</a>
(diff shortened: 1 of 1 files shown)


anselmbradford:
> Seems fine, but did you consider just removing the href attribute? The desired row hover effect still works if you do that.

The functionality of this appears to be around making the link click work for the whole row, so it's not just about the hover. See https://github.com/cfpb/capital-framework/blob/canary/src/cf-tables/src/TableRowLinks.js#L34

PR #660, opened 2017-10-18
Update JS to ES6

anselmbradford:
## Changes

- Update cf-tables and cf-expandables to ES6 syntax.
- Remove "Organism" suffix from table and expandable names (the type should be able to be looked up through the prototype, also this is obfuscated through minification anyway).
- Prefer named methods vs anonymous functions.

## Testing

- https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-components-locally expandables and tables should still work like the live site.

### DIFF
--- a/src/cf-expandables/src/Expandable.js
+++ b/src/cf-expandables/src/Expandable.js
@@ -4,23 +4,27 @@
 
 'use strict';
 
-var contains = require( 'cf-atomic-component/src/utilities/dom-class-list' ).contains;
-var addClass = require( 'cf-atomic-component/src/utilities/dom-class-list' ).addClass;
-var closest = require( 'cf-atomic-component/src/utilities/dom-closest' ).closest;
-var removeClass = require( 'cf-atomic-component/src/utilities/dom-class-list' ).removeClass;
-var ExpandableTransition = require( 'cf-atomic-component/src/utilities/transition/ExpandableTransition' );
-var Events = require( 'cf-atomic-component/src/mixins/Events.js' );
-var Organism = require( 'cf-atomic-component/src/components/Organism' );
+const domClassList = require( 'cf-atomic-component/src/utilities/dom-class-list' );
+const addClass = domClassList.addClass;
+const contains = domClassList.contains;
+const removeClass = domClassList.removeClass;
+const closest = require( 'cf-atomic-component/src/utilities/dom-closest' ).closest;
+const ExpandableTransition = require( 'cf-atomic-component/src/utilities/transition/ExpandableTransition' );
+const Events = require( 'cf-atomic-component/src/mixins/Events.js' );
+const Organism = require( 'cf-atomic-component/src/components/Organism' );
 
-var ExpandableOrganism = Organism.extend( {
+const Expandable = Organism.extend( {
+  ui: {
+    base:    '.o-expandable',
+    target:  '.o-expandable_target',
+    content: '.o-expandable_content',
+    header:  '.o-expandable_header'
+  },
 
-  transition: null,
-  accordionEvent: null,
-  activeAccordion: false,
   classes: {
-    targetExpanded: 'o-expandable_target__expanded',
+    targetExpanded:  'o-expandable_target__expanded',
     targetCollapsed: 'o-expandable_target__collapsed',
-    groupAccordion: 'o-expandable-group__accordion'
+    groupAccordion:  'o-expandable-group__accordion'
   },
 
   events: {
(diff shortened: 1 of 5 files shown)


src/cf-tables/src/TableRowLinks.js
```diff
@@ -1,38 +1,38 @@
...
+ */
+function onRowLinkClick( event ) {
+  let target = event.target;
+  if ( target && target.tagName === 'A' ) {
+    return;
   }
-};
+  target = closest( event.target, 'tr' );
+  const link = target.querySelector( 'a' );
+  if ( link ) window.location = link.getAttribute( 'href' );
```
jimmynotjim: This should be wrapped in curly braces to match the return on L:31

src/cf-tables/src/TableSortable.js
```diff
@@ -2,211 +2,215 @@
...
-
-    for ( var i = 0, len = rows.length; i < len; ++i ) {
-      cell = rows[i].cells[columnIndex];
-      if ( cell ) {
-        cell = cell.textContent.trim();
+/**
+ * Function used to create computed and triggered properties.
+ */
+function initialize() {
+  this.sortClass = UNDEFINED;
```
jimmynotjim: Should these be part of a `customClasses` object like in the Expandables code instead of properties?

src/cf-expandables/src/Expandable.js
```diff
@@ -4,87 +4,110 @@
...
-      addClass( this.ui.target, this.classes.targetExpanded );
-      removeClass( this.ui.target, this.classes.targetCollapsed );
-    }
+/**
+ * Initialize a new expandable.
+ */
+function initialize() {
+  const customClasses = {
+    BASE_CLASS:   'o-expandable_content__transition',
+    EXPANDED:     'o-expandable_content__expanded',
```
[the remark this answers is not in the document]
anselmbradford: Hmm, that's a good point! Let's revisit that in https://github.com/cfpb/capital-framework/issues/665, lest I break some existing functionality here.

src/cf-expandables/src/Expandable.js
```diff
@@ -4,87 +4,110 @@
...
+    COLLAPSED:    'o-expandable_content__collapsed',
+    OPEN_DEFAULT: 'o-expandable_content__onload-open'
+  };
+
+  const transition = new ExpandableTransition( this.ui.content, customClasses );
+  this.transition = transition.init();
+
+  const groupElement = closest( this.ui.target, '.' + this.classes.groupAccordion );
```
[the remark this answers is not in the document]
anselmbradford: Totally! I think everything would move to the group so we could eliminate the `closest` helper. (Related issue https://github.com/cfpb/capital-framework/issues/665)

src/cf-tables/src/TableSortable.js
anselmbradford: This looks like a different use case. In Expandables there's a set of class names that are assigned to constants that are only used inside the `initialize` method, so their scope is fine. In this case, `sortClass` is not constant and is used across methods, so is assigned as a property to share between methods. Personally I prefer not to expose anything publicly that is not needed in the public API, lest it leads to unintended implementations, so if `sortClass` could be internal that would be better (such as by moving it's location above the scope of the TableSortable object), but I don't think that's currently possible here as the methods here get added to instances of the Table organism, so that if `sortClass` was in a higher scope it would then become shared among all Table organisms using the TableSortable modifier. Is that right @sebworks?

src/cf-expandables/src/Expandable.js
```diff
@@ -4,87 +4,110 @@
...
-      this.activeAccordion = false;
-      this.transition.collapse();
-    }
-  },
-
-  onExpandableClick: function( event ) {
-    this.transition.toggleExpandable();
-    this.toggleTargetState( this.ui.target );
-  },
+  transition: null,
```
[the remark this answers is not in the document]
anselmbradford: Fixed

src/cf-expandables/src/Expandable.js
```diff
@@ -4,87 +4,110 @@
...
-var ExpandableTransition = require( 'cf-atomic-component/src/utilities/transition/ExpandableTransition' );
-var Events = require( 'cf-atomic-component/src/mixins/Events.js' );
-var Organism = require( 'cf-atomic-component/src/components/Organism' );
-
-var ExpandableOrganism = Organism.extend( {
+const domClassList = require( 'cf-atomic-component/src/utilities/dom-class-list' );
+const contains = domClassList.contains;
+const addClass = domClassList.addClass;
+const closest = require( 'cf-atomic-component/src/utilities/dom-closest' ).closest;
+const removeClass = domClassList.removeClass;
```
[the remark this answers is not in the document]
anselmbradford: Fixed

src/cf-tables/src/TableRowLinks.js
anselmbradford: Fixed

src/cf-tables/src/TableSortable.js
sebworks: yep, that is correct @anselmbradford

anselmbradford: Hmm, sidenote that my renaming these capitalized like they are classes when they're not instantiating new instances is possibly misleading.

PR #662, opened 2017-10-19
Remove on-ready check

anselmbradford:
On-ready check ensures `element` is not undefined, but `BaseTransition` [reference to `element`](https://github.com/cfpb/capital-framework/blob/canary/src/cf-atomic-component/src/utilities/transition/ExpandableTransition.js#L32) is not wrapped in on-ready, so if `element` is undefined, there will be greater problems than where it is being checked. 

If we want an on-ready function as a precaution in case CF is not added in footer, we could add on-ready at a higher level within the code, as opposed to within individual components.

## Removals

- Remove on-ready check in ExpandableTransition.

## Testing

- https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-components-locally expandable docs should be unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 -
 
 ### Removed
--
+- **cf-expandables:** [PATCH] Remove on-ready check.
 
 
 ## 4.14.1 - 2017-10-17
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 -
 
 ### Removed
--
+- **cf-expandables:** Remove on-ready check.
```
jimmynotjim: Don't forget `[MINOR]` here

anselmbradford: `[PATCH]`, right?

jimmynotjim: Yeah, that's prob better.

PR #663, opened 2017-10-19
Reference cf-atomic-component locally

anselmbradford:
Now that cf-atomic-component is apart of CF, we don't need to reach into node_modules for it, but instead can reference it through relative paths.

## Changes

- References cf-atomic-component requires locally.

## Testing

- https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-components-locally expandable docs should be unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 ### Changed
 - **cf-expandables:** [PATCH] Update to ES6 syntax.
 - **cf-tables:** [PATCH] Update to ES6 syntax.
+- **cf-atomic-component:** [PATCH] Reference cf-atomic-component locally.
 
 ### Removed
 - **cf-expandables:** [PATCH] Remove on-ready check.
--- a/src/cf-atomic-component/src/utilities/transition/ExpandableTransition.js
+++ b/src/cf-atomic-component/src/utilities/transition/ExpandableTransition.js
@@ -1,10 +1,10 @@
 'use strict';
 
 // Required modules.
-const Events = require( 'cf-atomic-component/src/mixins/Events.js' );
-const BaseTransition = require( 'cf-atomic-component/src/utilities/transition/BaseTransition' );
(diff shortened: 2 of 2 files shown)


Scotchester:
The change makes sense in `src/cf-atomic-component/src/utilities/transition/ExpandableTransition.js`, but in the other components, wouldn't it be more "standard" to keep the root of the `require` as the name of the module?

anselmbradford:
> The change makes sense in src/cf-atomic-component/src/utilities/transition/ExpandableTransition.js, but in the other components, wouldn't it be more "standard" to keep the root of the require as the name of the module?

Hmmm, so it references the one in `node_modules`, e.g. `capital-framework/src/cf-expandables/node_modules/cf-atomic-component`, which means if you're editing `capital-framework/src/cf-atomic-component` you won't see the changes in the docs for cf-expandables till the next release. But you bring up a good point, if a user installed only cf-expandables and referenced the source it would break. Maybe we don't care about that? Maybe they should only be referencing the minified JS? Do you see a better way where we can edit `capital-framework/src/cf-atomic-component` and have the changes show up in the dependent components locally?

jimmynotjim:
>  Do you see a better way where we can edit capital-framework/src/cf-atomic-component and have the changes show up in the dependent components locally?

`npm link` is probably the best way

Scotchester:
Yeah, what Jimmy said. Maybe I'm not understanding you, @anselmbradford, but wouldn't linking it into a separate folder where you're testing the docs site (which is our recommended way to work on docs right now) work?

anselmbradford:
> Yeah, what Jimmy said. Maybe I'm not understanding you, @anselmbradford, but wouldn't linking it into a separate folder where you're testing the docs site (which is our recommended way to work on docs right now) work?

I'm not certain it would, if I linked cf-expandables and cf-atomic-component from the main capital-framework repo, cf-expandables would still reach into its own node_modules for cf-atomic-component, wouldn't it. I would separately need to go into cf-expandables and do a linking in there. 

Anyway, the changes can be manually moved worst case, so that's fine for now. I rebased this PR to just be the cf-atomic-component references.

Issue #669, opened 2017-10-19
Make ExpandableTransition immutable


anselmbradford:
ExpandableTransition currently allows [overriding of the transition classes](https://github.com/cfpb/capital-framework/blob/canary/src/cf-atomic-component/src/utilities/transition/ExpandableTransition.js#L30), which means `u-expandable-transition` would potentially not be consistently implemented for this type of transition. My perspective is the transitions should be immutable. They should describe a set of concrete CSS classes that are associated with an API specific to the transition (e.g. `moveRight` for a `MoveTransition` and `fadeIn` for an `AlphaTransition`). The `BaseTransition` should be the only module that can have its associated classes overridden and it should never be instantiated directly, but instead should appear through the other transitions through composition.

I propose creating a concrete transition `MaxHeightTransition` or similar that describes what `ExpandableTransition` is doing and move `ExpandableTransition` to the cf-expandables package, where it can composite on top of `MaxHeightTransition` and add its own CSS classes.

## Current behavior

- ExpandableTransition has mutable CSS classes.

## Expected behavior

- Transitions should have a set of CSS classes that are immutable.

PR #670, opened 2017-10-19
Add transition tests

anselmbradford:
Add transition tests from https://github.com/cfpb/cfgov-refresh/pull/3493

## Additions

- Adds unit test coverage for BaseTransition, MoveTransition, and AlphaTransition.

## Testing

- `gulp test:unit` should have increased coverage.

## Todos

- ExpandableTransition doesn't have a test, but https://github.com/cfpb/capital-framework/issues/669 should be reconciled first.

### DIFF
--- a/test/unit-test/cf-atomic-component/utilities/transition/AlphaTransition-spec.js
+++ b/test/unit-test/cf-atomic-component/utilities/transition/AlphaTransition-spec.js
@@ -1,0 +1,57 @@
+'use strict';
+
+const srcPath = require( '../../src-path' );
+
+const chai = require( 'chai' );
+const expect = chai.expect;
+
+const AlphaTransition =
+  require( '../' + srcPath + 'utilities/transition/AlphaTransition' );
+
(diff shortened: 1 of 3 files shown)


test/unit-test/cf-atomic-component/utilities/transition/AlphaTransition-spec.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
+
+const srcPath = require( '../../src-path' );
```
jimmynotjim: I know this is merged, but `srcPath` is only ever used once in each of these files, do we need to be defining it?

jimmynotjim: NVM, I didn't realize exactly what this was doing till I looked at the other PR that created it.

anselmbradford: Yeah, just a convenience helper for the test paths.

PR #672, opened 2017-11-02
Self-host fonts

anselmbradford:
## Changes

- ~~Switches to [practical level of browser support](https://css-tricks.com/snippets/css/using-font-face/#article-header-id-1) to reduce font file numbers.~~ Pending internal discussion.
- Creates variable `@cf-fonts-path` for specifying where to load Avenir from.
- Updates usage docs.

## Testing

- ~~You'll need to add `@cf-fonts-path: 'http://www.consumerfinance.gov/fonts';` inside `src/static/css/main.less` for the docs site.~~
- ~~Visit any page in the docs and look at the path the fonts are coming from in the inspector. There should only be woff and woff2 fonts.~~

Okay, new testing instructions. Put this in `main.less` in gh-pages branch:

```less
// Webfont variables
// This is the path for self-hosted fonts.
@cf-fonts-path: '/static/fonts';
// Whether or not to override the local path and retrieve fonts from fonts.net.
@use-font-net: true;
```

Avenir should load locally and show woff2 being downloaded in the dev tools network tab.

Change it to:

```less
// Webfont variables
// This is the path for self-hosted fonts.
@cf-fonts-path: '/static/fonts';
// Whether or not to override the local path and retrieve fonts from fonts.net.
@use-font-net: false;
```

And Avenir should not load and you should see requests 404 for the fonts on localhost.


rosskarchner:
The only thing that jumps out at me:

maybe 

`@cf-fonts-path: 'http://www.consumerfinance.gov/fonts'; inside src/static/css/main.less` 

needs to be 

`@cf-fonts-path: 'https://www.consumerfinance.gov/fonts'; inside src/static/css/main.less` ?

anselmbradford:
@rosskarchner Scott and I tested on Friday, when we do http it redirects to https and cancels the request, so we get twice as many requests in the log. If we do direct https it straight away gets canceled.

rosskarchner:
What happens in other browsers? Is there anything in the javascript console?

rosskarchner:
Via chat, @anselmbradford showed me that there were CORS errors:

screen shot 2017-11-06 at 12 19 17 pm

rosskarchner:
I'd like to propose a different approach to all of this:

- We make a change in cfgov-refresh such that, if it is checked out beside cfgov-refresh, the contents of GHE/CFGOV/cfgov-fonts are added to Django's [STATICFILES_DIRS](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-STATICFILES_DIRS). 
- For developers, this would mean fonts are now available under /static/fonts instead of /fonts. CSS files should then be change to use relative references (since they will also be under /static)
- At deployment time, fonts would be gathered up during 'collectstatic', and given unique hashes, and served with long-term caching headers, the same as other static assets. (ie, the same mechamism that makes logo_sm-exec.svg available as /static/img/logo_sm-exec.9f5e268a182c.svg )

We would then need to move the Akamai configuration that protects /fonts to /static/fonts, but that's pretty simple.

rosskarchner:
Note that this is Django specific, and doesn't really help with CF's documentation site. Perhaps that could continue to use fonts.com?

anselmbradford:
@Scotchester What do you think about the above?

jimmynotjim:
I like @rosskarchner's idea, using fast.fonts.net to load the fonts outside of production is fine as long as we take that into consideration when doing speed tests.

anselmbradford:
Added variable to support pulling from fonts.net

src/cf-typography/src/cf-typography.less
```diff
@@ -13,6 +13,11 @@
 @webfont-medium: 'AvenirNextLTW01-Medium';
 @webfont-demi: 'AvenirNextLTW01-Demi';
 
+// Webfont variables
+@cf-fonts-path: '/fonts';
```
Scotchester: Let's change this to `/static/fonts` to match the config Ross set up over in cfpb/cfgov-refresh#3569.

src/cf-typography/src/licensed-fonts.less
```diff
@@ -0,0 +1,34 @@
+/*
+ * Source: http://fast.fonts.net/cssapi/44e8c964-4684-44c6-a6e3-3f3da8787b50.css
+ * This file has been edited to use absolute URLS so we can concatenate it with
+ * all of our other styles.
+ */
+@import url(//fast.fonts.net/t/1.css?apiType=css&projectid=44e8c964-4684-44c6-a6e3-3f3da8787b50);
+@font-face {
+    font-family: 'AvenirNextLTW01-Regular';
+    src: url( '@{cf-fonts-path}/1e9892c0-6927-4412-9874-1b82801ba47a.woff@{cf-fonts-path-suffix}' ) format( 'woff' ),
+         url( '@{cf-fonts-path}/2cd55546-ec00-4af9-aeca-4a3cd186da53.woff2@{cf-fonts-path-suffix}' ) format( 'woff2' );
```
Scotchester: WOFF2 should come before WOFF, right?

anselmbradford: Good catch! Fixed!

src/cf-typography/src/cf-typography.less
anselmbradford: Fixed

anselmbradford:
@Scotchester I re-added the legacy fonts so [GHE]/CFPB/hubcap/issues/66 doesn't need to hold this up for testing in prod.

anselmbradford:
Okay, new testing instructions. Put this in `main.less` in gh-pages branch:
```
// Webfont variables
// This is the path for self-hosted fonts.
@cf-fonts-path: '/static/fonts';
// Whether or not to override the local path and retrieve fonts from fonts.net.
@use-font-net: true;
```

Avenir should load locally and show woff2 being downloaded in the dev tools network tab.

Change it to:

```
// Webfont variables
// This is the path for self-hosted fonts.
@cf-fonts-path: '/static/fonts';
// Whether or not to override the local path and retrieve fonts from fonts.net.
@use-font-net: false;
```

And Avenir should not load and you should see requests 404 for the fonts on localhost.

src/cf-typography/usage.md
```diff
@@ -49,6 +50,22 @@ project by duplicating the variable `@key: value`.
...
+
+```
+@cf-fonts-path: '/fonts'
+```
+
+If you want to load fonts from fonts.net, set the following variable to `true`,
+otherwise set it to `false` to use the self-hosted font path:
+
+```
+@use-font-net: true;
```
Scotchester: How about we rename this variable to `@use-font-cdn` to make it agnostic, in case we change providers in the future?

src/cf-typography/usage.md
```diff
@@ -49,6 +50,22 @@ project by duplicating the variable `@key: value`.
...
+### Font source variables
+
+Use this variable to specify where the fonts files used in the `@webfont-*`
+variables are located. Can be either a relative or absolute path.
+
+```
+@cf-fonts-path: '/fonts'
+```
+
+If you want to load fonts from fonts.net, set the following variable to `true`,
```
Scotchester: Let's change "fonts.net" here to "a Fonts.com Web Fonts project"

src/cf-typography/src/licensed-fonts.less
```diff
@@ -0,0 +1,103 @@
+/*
+ * Source: http://fast.fonts.net/cssapi/44e8c964-4684-44c6-a6e3-3f3da8787b50.css
+ * This file has been edited to use absolute URLS so we can concatenate it with
+ * all of our other styles.
+ */
+@import url(//fast.fonts.net/t/1.css?apiType=css&projectid=44e8c964-4684-44c6-a6e3-3f3da8787b50);
+
+.use-font-net( @use-font-net ) when ( @use-font-net = 'true' ) {
```
Scotchester: I think this mixin would be more appropriately titled something like `generate-font-face-rules`.

src/cf-typography/src/licensed-fonts.less
```diff
@@ -0,0 +1,103 @@
+/*
+ * Source: http://fast.fonts.net/cssapi/44e8c964-4684-44c6-a6e3-3f3da8787b50.css
+ * This file has been edited to use absolute URLS so we can concatenate it with
+ * all of our other styles.
+ */
+@import url(//fast.fonts.net/t/1.css?apiType=css&projectid=44e8c964-4684-44c6-a6e3-3f3da8787b50);
```
Scotchester: Let's add a comment for this line to indicate its purpose and that it should not be removed. How about:

```less
// Pings Fonts.com for pageview tracking. Must be used even when self-hosting the fonts.
```

src/cf-typography/src/licensed-fonts.less
```diff
@@ -0,0 +1,103 @@
+/*
+ * Source: http://fast.fonts.net/cssapi/44e8c964-4684-44c6-a6e3-3f3da8787b50.css
+ * This file has been edited to use absolute URLS so we can concatenate it with
+ * all of our other styles.
+ */
```
Scotchester: I think this whole comment can be removed, but add some kind of general header banner for the file.

Scotchester:
Oops, forgot to say that this does work as expected! Just had a few comments on meta stuff.

cfarm:
@anselmbradford do new testing instructions include this from original post?

> You'll need to add @cf-fonts-path: 'http://www.consumerfinance.gov/fonts'; inside src/static/css/main.less for the docs site.

if just follow your latest instructions, Avenir still loads from fastfonts.net with `@use-font-net: false`.

src/cf-typography/src/licensed-fonts.less
anselmbradford: Presume this should be fonts.net.

src/cf-typography/src/cf-typography.less
```diff
@@ -13,6 +13,12 @@
 @webfont-medium: 'AvenirNextLTW01-Medium';
 @webfont-demi: 'AvenirNextLTW01-Demi';
 
+// Webfont variables
+// This is the path for self-hosted fonts.
+@cf-fonts-path: '/static/fonts';
+// Whether or not to override the local path and retrieve fonts from fonts.net.
+@use-font-net: false;
```
Scotchester: Missed updating the variable here.

src/cf-typography/src/licensed-fonts.less
```diff
@@ -0,0 +1,105 @@
+/* ==========================================================================
+   Capital Framework
+   Licensed font family URLs.
```
Scotchester: How about "Licensed font URLs – for CFPB use only"?

cfarm:
Still having the same issue. I think the problem is with the `gh-pages` branch, which still imports licensed-fonts.css and sets the variables to use it (see PR #607). I don't think it is supposed to have a licensed-fonts.css file of its own.

Scotchester:
@cfarm Yeah, I saw that and had to remove that import. Forgot to mention it here.

cfarm:
Groovy. That also reminds me, we should declare those font variables in cf-typography (not downstream Less files), so I'll do that as part of the loading-fewer-Avenir-Next-weights work.

anselmbradford:
> Approved pending a companion PR with updates to the gh-pages branch to make sure the fonts load correctly there as expected.

I'll release and make a PR for gh-pages now.

anselmbradford:
This is in the docs now https://cfpb.github.io/capital-framework/components/

PR #673, opened 2017-11-08
Update dependences and build

anselmbradford:
## Changes

- Updates build gulp tasks to ES6 syntax.
- Replaces uglifier with `uglifyjs-webpack-plugin` to match cfgov-refresh configuration.
- Updates `babel-preset-env`, `deepmerge`, `finalhandler`, `run-sequence`, `serve-static`, `sinon`, `wcag`, `webpack`.

## Testing

- `gulp build && gulp test` should pass.


scripts/gulp/scripts.js
```diff
@@ -7,7 +7,24 @@ const gulpIgnore = require( 'gulp-ignore' );
 const gulpRename = require( 'gulp-rename' );
 const webpack = require( 'webpack' );
 const webpackStream = require( 'webpack-stream' );
-const vinylNamed = require('vinyl-named');
+const UglifyWebpackPlugin = require( 'uglifyjs-webpack-plugin' );
+const vinylNamed = require( 'vinyl-named' );
```
jimmynotjim: Do you need this plugin?

anselmbradford: It's used https://github.com/cfpb/capital-framework/commit/2681c26429f5b1b76175192c636e6b29193c75dc#diff-66f9adfefaa465f6da5be9394caa2259R34 @contolini ?

contolini: Yes, I believe it's used to name the files `cf-whatever.js`. See https://github.com/shama/vinyl-named#example

jimmynotjim: Ahhh, right. Carry on.

PR #677, opened 2017-11-16
Fix run-sequence so default Gulp task works again

Scotchester:
The default Gulp task (i.e., just running `gulp`) wasn't working because `build` was prematurely reporting completion and `test` was running before it should.

The fix is to restore the callback parameter to the `default` and `build` tasks and pass that to `runSequence` after the sequence of tasks.

I paired with @jimmynotjim on fixing this.

## Testing

1. Before pulling this branch, run `gulp` and see that it fails.
1. Pull this branch
1. Run `gulp` and see that it works

## Checklist

### DIFF
--- a/scripts/gulp/build.js
+++ b/scripts/gulp/build.js
@@ -3,11 +3,12 @@
 const gulp = require( 'gulp' );
 const runSequence = require( 'run-sequence' );
 
-gulp.task('build', () => {
+gulp.task( 'build', callback => {
   runSequence(
-    ['styles:cf', 'scripts:cf', 'clean:tmp'],
-    ['template:readmes', 'copy:components:boilerplate'],
-    ['copy:components:source', 'template:usage', 'copy:components:manifest'],
-    ['styles:components', 'scripts:components', 'styles:grid']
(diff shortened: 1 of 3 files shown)


scripts/gulp/build.js
```diff
@@ -3,11 +3,12 @@
...
   runSequence(
-    ['styles:cf', 'scripts:cf', 'clean:tmp'],
-    ['template:readmes', 'copy:components:boilerplate'],
-    ['copy:components:source', 'template:usage', 'copy:components:manifest'],
-    ['styles:components', 'scripts:components', 'styles:grid']
+    [ 'styles:cf', 'scripts:cf', 'clean:tmp' ],
+    [ 'template:readmes', 'copy:components:boilerplate' ],
+    [ 'copy:components:source', 'template:usage', 'copy:components:manifest' ],
+    [ 'styles:components', 'scripts:components', 'styles:grid' ],
+    callback
```
anselmbradford: You don't have to call the callback, ala `callback()`?

anselmbradford: Oh, wait, nevermind, I just realized this is being passed as an argument.

PR #691, opened 2017-11-29
Add font-display:swap to @font-face syntax

ascott1:
This adds `font-display: swap` to the `@font-face` syntax, which will create a flash of unstyled text (FOUT) in [supported browsers](https://caniuse.com/#feat=css-font-rendering-controls), rather than the default flash of invisible text (FOIT) behavior. Currently this is limited to Chrome, which is a high percentage of our users and will have no impact on non-Chrome users.

In the future if we would like to replicate this behavior in all browsers (using something like [Font Face Observer](https://fontfaceobserver.com/)), we _could_ check for support:

```js
if ("fontDisplay" in document.body.style === false) {
  // js for FOUT
}
```

## Additions

- `font-display: swap` property to all `@font-face` rules


Scotchester:
Cool idea.

What would you think about choosing one of the values for that property that allows a "small" or "extremely small" period of invisible text before resorting to the fallback font if the webfont hasn't yet loaded? From [the descriptions of the options on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display), I think I'd be in favor of trying `fallback`.

ascott1:
I'm open to `fallback`, but from my understanding of it, `swap` seems to be the recommended value from a performance standpoint. Zach Leatherman [recommends `swap`](https://www.zachleat.com/web/comprehensive-webfonts/#font-display) as does this [CSS-Tricks](https://css-tricks.com/font-display-masses/#article-header-id-2) article which states:

> If you don't know which option to use, then go with swap. Not only does it provide an optimal balance between custom fonts and accessibility of content, it provides the same font loading behavior that we've relied on JavaScript for. **If you have fonts on the page that you'd like to have load, but could ultimately do without, consider going with fallback or optional when using font-display.**

However this [Google Developer's guide](https://developers.google.com/web/updates/2016/02/font-display) states:

> fallback gives the font face an extremely small block period (100ms or less is recommended in most cases) and a short swap period (three seconds is recommended in most cases). In other words, the font face is rendered with a fallback at first if it’s not loaded, but the font is swapped as soon as it loads. However, if too much time passes, the fallback will be used for the rest of the page’s lifetime. **fallback is a good candidate for things like body text where you’d like the user to start reading as soon as possible and don’t want to disturb their experience by shifting text around as a new font loads in.**

[This pen](https://codepen.io/vio/full/dpNNYm/) shows pretty similar behavior between the two for me, when simulating a 3G connection in my browser. Ultimately, I think either are preferable to the default behavior. Happy to leave it up to the crowd and I can update the PR if we decide to go with `fallback`.

jimmynotjim:
> However, if too much time passes, the fallback will be used for the rest of the page’s lifetime

I think this is an important distinction. If the connection is really slow, wouldn't it be preferable to not flash at all until the page is reloaded or user navigates to a new page?

ascott1:
That does seem like an important point @jimmynotjim. Should I switch this PR to use `fallback` instead? Any reason not to do that?

Scotchester:
Yeah, I'm in favor of that. Might be good to get at least one more opinion in here, though. @cfpb/front-end-team-admin

anselmbradford:
This table is great for visualizing them https://css-tricks.com/almanac/properties/f/font-display/#article-header-id-0, as is the graphics here https://font-display.glitch.me/

`fallback` sounds good to me. Though that last article also recommends `optional` as it will allow the browser to decide based on the connection speed. We could use a media query to deliver `optional` to screensizes that are presumably mobile.

Scotchester:
@anselmbradford That's really interesting. The MDN docs I read earlier don't mention that aspect of `optional`.

However, I don't think we can media-query it, since it's a property of the `@font-face` declarations. Regardless, I'm worried it would result in too many instances of font downloads not making that 100ms cutoff and then the page is stuck on the fallback. I think I still vote for `fallback`.

anselmbradford:
> However, I don't think we can media-query it, since it's a property of the @font-face declarations. 

Ah, right. I was thinking of a variation of "Selectively loading custom fonts based on screen width" mentioned here https://developer.akamai.com/blog/2017/07/25/performance-usage-implications-custom-fonts/. That technique is to wrap the `font-family` in a media query and only ever deliver the custom font above a certain screen width.

anselmbradford:
Ooo @Scotchester thought of how we could actually do media queries and the font-face declaration. Using [`@import`](https://developer.mozilla.org/en-US/docs/Web/CSS/@import) we could import a font-face declaration file based on the media and have a separate font-face declaration for another type of media. Anyway, not something to pursue under this PR, but it likely fits under https://[GHE]/CFGOV/platform/issues/1832

ascott1:
I updated the rules to use `fallback` and updated the changelog as discussed.

cfarm:
Thanks for this @ascott1, I'm excited to see how this compares to the font loading api prototype we've been working on.

I want to demo this with @nataliafitzgerald at our fonts discussion today before we decide to merge. I can deploy this to a dev server first to share with her, so we can get an understanding of the performance and visual impact on cfgov-refresh.

ascott1:
> Thanks for this @ascott1, I'm excited to see how this compares to the font loading api prototype we've been working on.

This is a "light touch" version of that and currently only improves performance for Chrome users.

> I want to demo this with @nataliafitzgerald at our fonts discussion today before we decide to merge. I can deploy this to a dev server first to share with her, so we can get an understanding of the performance and visual impact on cfgov-refresh.

This would be great @cfarm. Thank you!

ascott1:
@cfarm did you have a chance to get design feedback on this? Does it require any further discussion?

cfarm:
I got caught up with TDP this week, working on the deploy now.

cfarm:
Merging to test this out on cfgov-refresh! We currently can't test there without releasing an individual component, due to issue described #708

PR #693, opened 2017-11-29
Docs: Fix npm badges for cf-tables and cf-typography

Scotchester:
SSIA


Scotchester:
Since those lines were added over a year and a half ago :)

PR #697, opened 2017-12-01
Remove older font formats

anselmbradford:
## Removals

- Removes all but woff/woff2 support.

## Testing

- Follow local testing directions.
- Open IE9 in a VM. Change to IE8 browser mode.
- Capture a network profile and see that the fonts don't load.

## Screenshots

screen shot 2017-12-01 at 5 41 46 pm


anselmbradford:
Updated to major semver version based on @jimmynotjim feedback.

Scotchester:
Not sure I agree with that. What is your reasoning, @jimmynotjim?

jimmynotjim:
@Scotchester it's removing an existing functionality. I know internally we've weighed that choice, and are ok with it, but if we think about it from an outside perspective, it's a breaking change.

Scotchester:
Given the miniscule number of users it will impact, I'd argue that it's not "breaking".

jimmynotjim:
Does size matter? If only a single page that's rarely visited used something like the Hero, we'd still release a major version if we removed it. A breaking change is a breaking change right?

Scotchester:
I definitely think the definition of breaking is fuzzy and sometimes a judgment call. I think it depends on the magnitude of the change. In this case, "breaking" just means that some users won't see the preferred typeface. That may be, by a very strict definition, "breakage" from past behavior, but the tiny number of users it will affect also likely won't even notice. Or they'll notice that the site performs better and be happier, anyway.

From a practical point of view, the notion of declaring a major/breaking change is to signal to a developer that if they don't make some updates to their project's code, it will not look/behave as before after the update. While this is strictly true, I can't imagine any developer in almost-2018 making the effort to re-add support for these older font formats.

All that said, I'm willing to be overridden on this. Any other opinions?

jimmynotjim:
That's the part of semver that's super confusing for a project like this. There is no API, so are we versioning based on the visitor's experience or the developer's? I'm fine if we say the developer's, but I do think we need to define it and share it across the team.

jimmynotjim:
Realizing I've sidetracked this PR quite a bit, moving the semver discussion to a new issue. @anselmbradford feel free to revert the major changelog entry and merge when you're ready.

PR #714, opened 2017-12-13
Remove Avenir Next italic font 

cfarm:
Removes Avenir Next italic font face to reduce the number of web font files downloaded by users. On pages that currently load the Italic font file, Avenir Next Regular will be used instead, and the browser will italicize the font for us.

## Removals

- remove `font-face` rules for the Avenir Next italic web font

## Changes

- change font variable declarations from Avenir Next Italic to Avenir Next Regular. Pages using italic text will no longer load an additional font in order to render that text. They will use the regular font and the browser will apply slanted italic styling.

## Testing

- Check out this branch
- Follow directions here https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-components-locally
- Since there are no examples of italic text in the Capital Framework docs, if you want to view one, you'll have to add it manually. Add this text to the `cf-typography/usage.md` file, and hit save in your test repo's `components/cf-typography/index.md` file to view it locally at http://localhost:3000/components/cf-typography/:

```markdown
## Italic text

Lorem <em>ipsum dolor sit amet</em>, consectetur adipisicing elit, sed do eiusmod
tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
consequat. <em>Duis aute irure dolor in reprehenderit in voluptate velit esse
cillum dolore eu fugiat nulla pariatur</em>. Excepteur sint occaecat cupidatat non
proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
```

## Screenshots
screen shot 2017-12-13 at 11 06 10 am

Additional screenshots and comparisons of examples from cfgov-refresh are documented at https://GHE/CFGOV/platform/issues/2347#issuecomment-140554

## Notes on performance

- Pages that use the italic font currently will see a 19kb reduction in page weight once this is deployed.
- Pages with this reduction in page weight have shown an average decrease of 1.07 seconds in the first meaningful paint time, for desktop users on a broadband connection. For a simulated mobile connection, the average decrease for this metric was 3.19s. [See page load metrics on GHE](https://GHE/CFGOV/platform/issues/2347#issuecomment-140861).
- Few pages use the italic font style, so most pages will not see a performance gain. Key pages that do use the italic style include Owning a Home and Ask CFPB (on the Feedback form module) and the [Learn more](www.consumerfinance.gov/learnmore/) page, all of which receive significant traffic.

## Checklist

[...]

Issue #700, opened 2017-12-02
Use Avenir Next family instead of full name with style


anselmbradford:
## Current behavior

- We use font-family names like `AvenirNextLTW01-Regular` and `AvenirNextLTW01-Italic` when we could use a single family name.

## Expected behavior

- Follow the example in https://alistapart.com/article/using-webfonts to specify a single font-family name and set the styles as bold, italic, etc.

jimmynotjim:
💯 I've been rallying for this forever. Excited that it's now possible.

Scotchester:
Proposed steps:

1. Add new `@font-face` rules for font-family "Avenir Next".
   - Make sure to a[dd the `local` names](https://www.broken-links.com/2009/06/30/checking-for-installed-fonts-with-font-face-and-local/) for users that already have it installed. I'm not entirely sure what those names should be, but they should be discoverable somehow.
1. Replace [the mixin on the `body` element](https://github.com/cfpb/capital-framework/blob/canary/src/cf-core/src/cf-base.less#L61) with `font-family: "Avenir Next", Arial, sans-serif;`.
1. Update all Capital Framework Less to replace calls to the `.u-webfont-` mixins with the appropriate `font-style` or `font-weight` property.
1. Leave old `font-face` declarations and `.u-webfont-` mixins in place and bump cf-core with a **minor** update. This will allow some transition time for other projects before we remove the old method in a later major release.

jimmynotjim:
We could leave the mixin definitions but change where they are being used to the new `font-family` stack without a major release right?

Scotchester:
@jimmynotjim Not sure I follow. Example in code?

jimmynotjim:
Sorry, realizing you're # 3 is exactly what I was trying to say.

PR #719, opened 2017-12-21
HOLD: Update cf-typography docs with italic web font removal + font-display: fallback

cfarm:
Updates the docs after removal of Avenir Next Italic in #714 and addition of font-display font-face property in #691.

Also includes whatever other changes from latest release might affect CSS.


_includes/usage/cf-typography.md
```diff
@@ -45,7 +45,7 @@ project by duplicating the variable `@key: value`.
 
 ```
 @webfont-regular: 'AvenirNextLTW01-Regular';
-@webfont-italic: 'AvenirNextLTW01-Italic';
```
Scotchester: I think it's worth adding a note to the docs (above or below here) about _why_ we're doing this, noting the fact that (a) faux italic Avenir Next Regular is almost indistinguishable from Avenir Next Italic, and (b) we don't use a lot of italic as a matter of course in our websites.

anselmbradford: Might roll it into https://github.com/cfpb/capital-framework/issues/700 if that makes more sense.

Scotchester: Oy, just realized that this is the docs PR and it needs to be updated in the main branch's usage.md file first. Stand by.

Scotchester:
These changes got merged in via other means.

PR #723, opened 2017-12-29
Add gitbook docs

anselmbradford:
Sets up the docs in [GitBook](https://gitbook.com/), which removes ruby dependency of jekyll and adds search and the ability to link to docs for editing direct from the docs.

Adds /docs/ directory, which can [host the docs](https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/) on GH pages if this is merged into the master branch as well.

Docs are symlinked to the usage docs so they are used directly. They also pull in the current CF version into the docs header as set in package.json.

## Additions

- /docs/ directory to host GitBooks project.
- `gulp docs:build` and `gulp docs:start` commands.

## Changes

- Adjusts docs to fit with GitBooks markdown parsing.
- Uses prism for syntax highlighting.

## Testing

1. Run `npm install`
1. Run `gulp docs:build`
2. Run `gulp docs:start`
3. Visit the docs at http://localhost:4000
4. Edit usage.md docs in /src/cf-buttons or whatever to see that they live reload.
4. You can try changing the version set in `package.json` and re-running step 1-3 to see that it auto-pulls in the version number into the docs heading.

## Screenshots

2017-12-28-21-23-localhost_4000

Search example:
screen shot 2017-12-28 at 9 24 24 pm

## Notes

- `SUMMARY.md` is auto-generated by `auto-summary` plugin.
- Uses freshly created https://github.com/cfpb/gitbook-plugin-theme-cfpb, which we may be able to customize further to use the old doc layout. But I wanted to get an MVP in that eliminated the gh-pages clone requirement for docs.


anselmbradford:
We can write the symlinks to the docs using gulp 4 if we want https://github.com/gulpjs/gulp/blob/master/docs/API.md#gulpsymlinkfolder-options. Which would allow a complete clean/build sequence.

jimmynotjim:
> But I wanted to get an MVP in that eliminated the gh-pages clone requirement for docs.

This is exciting. I know we'll still have to do npm linking to test cfgov-refresh and the Design Manual, but it'll be great to eliminate that requirement here.

docs/book.json
```diff
@@ -0,0 +1,40 @@
+{
+  "gitbook": "3.x.x",
```
jimmynotjim: What's the version here used for? Should it be pinned?

src/cf-core/src/cf-core.less
```diff
@@ -3,8 +3,8 @@
    Core Less file
    ========================================================================== */
 
-@import (less) '../../normalize-css/normalize.css';
-@import (less) '../../normalize-legacy-addon/normalize-legacy-addon.css';
+@import (less) '../../../node_modules/normalize-css/normalize.css';
```
jimmynotjim: Does this path continue to work when this module is installed into other projects?

Scotchester: There may be a reason that this is like this originally that I don't remember, but if Less is configured correctly, these relative paths shouldn't be needed at all, right? Should just be able to do `@import (less) 'normalize-css/normalize.css';`.

src/cf-forms/usage.md
```diff
@@ -750,9 +771,7 @@ typically to clear the input.
 
 <div class="m-btn-inside-input">
     <input type="text"
-        value="This is some really long text to make sure that the button
-               doesn't overlap the content in such a way that this input
-               becomes unusable."
+        value="This is some really long text to make sure that the button doesn't overlap the content in such a way that this input becomes unusable."
```
jimmynotjim: Is it no longer possible to break this under 80char line lengths?

src/README.md
```diff
@@ -0,0 +1,66 @@
+# Introduction
```
jimmynotjim: Should all this content be moved to the main `README.md` file so that we don't have duplicate readmes?

docs/book.json
anselmbradford: It's for specifying the gitbook version for plugins https://toolchain.gitbook.com/config.html. Not sure if it can be pinned. It does accept things like ">= 3.0.0"

src/cf-forms/usage.md
anselmbradford: I don't think this ever worked correctly, check out https://cfpb.github.io/capital-framework/components/cf-forms/

The whitespace ends up in the output
screen shot 2018-01-02 at 2 05 45 pm

jimmynotjim: Oh, hahaha, glad you caught it

src/README.md
anselmbradford: I suppose that could work. Is there anything on the Readme that appears on GH that we don't want in the docs and vice versa?

src/cf-core/src/cf-core.less
jimmynotjim: Close, it's the build manager (in our case Gulp) that needs to be configured. Because we can't know for sure that the build task is set up to glob the node_modules directory, I don't think we can remove the full path.

Scotchester: > Because we can't know for sure that the build task is set up to glob the node_modules directory, I don't think we can remove the full path.

Ahhh, that's right. I know it's run by Gulp in this case, but yeah, this module can't rely on anything other than it being installed by npm.

But back to your original question, I'm not sure why this is necessary at all.

jimmynotjim:
RIP custom gitbooks themes. @anselmbradford we should close this yeah?

PR #724, opened 2018-01-02
Update to gulp 4

anselmbradford:
## Additions

- `ajv` as a required peer-dependency.

## Removals

- Removes `run-sequence` and replaces it with native `gulp.series` and `gulp.parallel`.
- Remove deprecated `gulp-rimraf` and use native `del` instead.

## Changes

- Updates gulp 3.9.1 to gulp 4.0.0.
- Replace `gulp-json-format` with `gulp-jsbeautifier`, to handle issue where gulp-json-format didn't handle the pipe coming in from gulp-rename.
- Updates `chalk`, `gulp-autoprefixer`, `gulp-istanbul`, `gulp-less`, `gulp-markdown`, `promisify-node`, `uglifyjs-webpack-plugin`.
- Moves gulp tasks below `build` and `default` to a subdirectory `/tasks/`. This is because the order of the task definitions in the gulpfile matters. build and default can exist together in the same directory because default comes last. If the name was zbuild, it would not work.
- Fixes issue where build scripts weren't being linted because path was wrong.
- Autofixes linter issues in build scripts.

## Testing

- `npm install && gulp build && gulp test && gulp test:unit && gulp lint` (lint will have existing errors)


jimmynotjim:
> This is because the order of the task definitions in the gulpfile matters. build and default can exist together in the same directory because default comes last. If the name was zbuild, it would not work.

This has got to be a bug right? Should we hold off upgrading till we're sure this is the intended behavior?

anselmbradford:
> This has got to be a bug right? Should we hold off upgrading till we're sure this is the intended behavior?

It is not a bug, it is how it works. [This thread](https://github.com/gulpjs/gulp/issues/802#issuecomment-66195378) has background.
I'm not totally clear why this worked before, but anyway.

As I see it we have three options:
- The way in this PR.
- Prefix the tasks with numbers so they are sucked up by `require-dir` in the desired order.
- Move `default` and `build` inline into the bottom of the gulpfile (since only those two have cross-references to other tasks that matter).

anselmbradford:
Also, they encourage named functions for task names instead of strings, but I think with our heavy use of `:`, strings are the way to go.

anselmbradford:
Oh, wait I overlooked the end of that thread. Looks like it can be done with https://github.com/gulpjs/undertaker-forward-reference too. Not sure it's worth an extra dependency though.

anselmbradford:
Closed in favor of https://github.com/cfpb/capital-framework/pull/737

PR #725, opened 2018-01-04
Move ignored autoprefixer rule to its own block

anselmbradford:
Autoprefixer ignore comment applies to the whole block, so wasn't being re-enabled.

## Changes

- Move ignored autoprefixer rule to its own block

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 ### Changed
 - **cf-forms:** [PATCH] Add transform to fix radio button Firefox rendering bug.
 - **cf-forms:** [MINOR] Add multi-line support to checkboxes/radio buttons.
+- **cf-forms:** [PATCH] Move autoprefixer ignored rule to its own block.
 
 ### Removed
 -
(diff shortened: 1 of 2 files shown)


src/cf-forms/src/molecules/form-fields.less
```diff
@@ -8,19 +8,22 @@
         margin-top: unit( 5px / @base-font-size-px, em );
     }
 
+    /*
+    We need to turn off autoprefixing for the inline-grid because
+    older IE does not handle an inline-grid like other browsers,
+    leading to an extremely narrow column of text for the label.
+     */
```
Scotchester: It's probably worth mentioning why this has to be in its own block.

anselmbradford: Done

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 ### Changed
 - **cf-forms:** [PATCH] Add transform to fix radio button Firefox rendering bug.
 - **cf-forms:** [MINOR] Add multi-line support to checkboxes/radio buttons.
+- **cf-forms:** [PATCH] Move autoprefixer ignored rule to its own block.
```
Scotchester: You can leave it if you want, but seeing as the thing we're correcting hasn't been released yet, this changelog update may be unnecessary?

anselmbradford: Eh I'll leave it. Doesn't hurt to single it out since it's in its own PR.

PR #726, opened 2018-01-05
Lints release and build scripts

anselmbradford:
## Changes

- Runs autolinting on release and build scripts.
- Restructures lint task to have defined methods for tasks.
- Fixes issue where build scripts weren't being linted because path was wrong.
- Replace `gulp-json-format` with `gulp-jsbeautifier`.
## Testing

- `npm install`
- `gulp lint:build && gulp lint:tests && gulp lint:scripts && gulp lint:release` should run and will have errors for the tests (which are deprecated) and for the release scripts, which will need to be incrementally updated in the future.

## Notes

- Spun out of https://github.com/cfpb/capital-framework/pull/724

PR #730, opened 2018-01-05
Re-structure lint tasks

anselmbradford:
## Changes

- Adds named functions for lint tasks instead of anonymous functions.
- Fixes incorrect path for build files in lint task.

## Testing

- `gulp lint` should pass.

## Notes

- Spun out of https://github.com/cfpb/capital-framework/pull/726


scripts/gulp/lint.js
```diff
@@ -22,53 +20,63 @@ function _genericLintJs( src ) {
...
 /**
  * Lints the test js files for errors.
+ * @returns {Object} An output stream from gulp.
  */
-gulp.task( 'lint:tests', () => {
-  _genericLintJs( [
+function lintTests() {
+  return _genericLintJs( [
     'test/accessibility/*.js',
     'test/*.js'
```
jimmynotjim: This could be simplified as `test/**/*.js`. Currently it's not catching any of the unit tests for atomic component

Scotchester:
Probably a dumb question, but: Why?

anselmbradford:
> Probably a dumb question, but: Why?

Named functions are generally better as they'll have a name in any call stack errors. It also allows us to group the task definitions together at the bottom, and it's the [gulp recommend approach](https://github.com/gulpjs/gulp/issues/802#issuecomment-66195466). Lastly, ESlint will complain about undocumented functions, but it won't complain about undocumented anonymous functions.

scripts/gulp/lint.js
anselmbradford: Done! Good catch

PR #732, opened 2018-01-05
Add support for family-linked webfonts

Scotchester:
Step one in tackling #700. Step two will be removing the deprecated mixins in the CFv5 release.

## Additions

- New `@font-face` rules for family-linked webfonts

## Changes

- Replace use of `u-webfont-` mixins with standard properties in all CF components

## Testing

1. Pull branch
1. Run `gulp` to see successful build and tests
1. [Test with the docs site](https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-in-the-documentation-site)
  - Note that you'll have to inspect something that isn't explicitly set to use one of the old mixins in the docs' Less, or wait to test with my corresponding docs PR that is forthcoming.

## Screenshots

screen shot 2018-01-05 at 17 08 43

## Notes

- I set these all as `[MINOR]`, since we're not deprecating the mixins until CFv5, however, it's a little fuzzy, since styles in CF components that previously relied on them could get faux italic or faux bold if a component is updated without _also_ updating cf-core and cf-typography. If folks strongly object to calling that a minor update, let me know. 

## Todos

- [x] Submit corresponding docs PR

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [x] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [x] Passes all existing automated tests
* [x] New functions include new tests
* [x] New functions are documented (with a description, list of inputs, and expected output)
* [x] Placeholder code is flagged
* [x] Visually tested in supported browsers and devices
* [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### Browsers

- [x] Chrome
- [x] Safari
- [x] FF
- [ ] ~~IE~~ _Note: IE was unable to be tested because I couldn't get Fonts.com to serve the fonts to an IE VM. It always returns 403 Forbidden, even after adding the IP address to the whitelist. We'll test it live on the docs site._

PR #733, opened 2018-01-05
Docs: Update to stop using webfont mixins

Scotchester:
Corresponding to #732, this update will remove all of the uses of the `u-webfont-` mixins from the docs site's Less and adds a deprecation notice to the cf-core page.

Looks like it also brings in a couple other minor updates from some components' usage files.

I also couldn't stop myself from making a few style fixes and tweaks while I was in there.


_includes/usage/cf-grid.md
```diff
@@ -36,16 +36,6 @@ Read on for more details on the variables and mixins this component provides.
   or your compiled Capital Framework CSS will
   not work perfectly in older browsers.
 
-## Table of contents
```
jimmynotjim: @Scotchester did you add this here? It seems the usage.md this is copied from doesn't include a ToC.

jimmynotjim:
Updated for the release of v4.21.0. Does anyone else from @cfpb/front-end-team-admin want to review this before I merge it?

_includes/usage/cf-grid.md
anselmbradford: I added it back in October https://github.com/cfpb/capital-framework/commit/ba782f45243b0e89b3e40916ff289b79c23070fc#diff-03a6d671db586ccdd7a5dd6650e1298c

I think it should stay, particularly since we have variables in a TOC elsewhere. The more similarly formatted the components usage.md files are, the better.

jimmynotjim: @anselmbradford Weird, why is it not included in the build when I run it but is when Scott runs it? Could you pull this branch down and give `npm install && gulp build` a try?

jimmynotjim:
@Scotchester confirmed the changes work in IE 11 on the live docs

screen shot 2018-01-12 at 11 42 07 am

Scotchester:
Woohoo!

PR #736, opened 2018-01-08
Support autofix flag in style lint

anselmbradford:
## Changes

- Updates to latest stylelint.
- Adds support for `--fix` flag.
- Ignores node_modules directories inside CF components.
- Convert stylelintrc to yaml to match eslintrc format.

## Testing

- `npm install && gulp lint:styles`
- `gulp lint:styles --fix` (there will be four files that are "fixed" by removing the comments, you'll want to revert those till we can figure out what rule configuration will support comments inside of less functions).


scripts/gulp/lint.js
```diff
@@ -49,12 +49,22 @@ gulp.task( 'lint:scripts', () => {
...
+  // Pass all command line flags to Stylelint.
+  const options = minimist( process.argv.slice( 2 ) );
+  const willFix = options.fix || false;
+  return gulp.src( [
+      '!src/cf-*/node_modules/**/*.less',
+      '!src/cf-grid/src-generated/*.less',
+      'src/**/*.less'
+    ] )
     .pipe( gulpStylelint( {
+      failAfterError: true,
```
jimmynotjim: Is this the bit that was being weird in our testing yesterday?

anselmbradford: Yeah you can tell it to fail on an error or not. I guess we want it to fail, as that would fail Travis.

jimmynotjim: Oh good point. Only question is why isn't Travis failing now since there are errors?

anselmbradford: That's a good question. We'd need `gulp lint` in [here](https://github.com/cfpb/capital-framework/blob/canary/.travis.yml), right?

jimmynotjim: Yep I think you're right. Now that I think about it, we also need to fix those errors in this PR after we've confirmed it stops Travis.

anselmbradford: Can be another PR. This just added the autofix flag, which Travis won't be using.

anselmbradford: Eh nevermind, I'll push it to this PR.

anselmbradford: You can see the failure in https://travis-ci.org/cfpb/capital-framework/builds/327020639?utm_source=github_status&utm_medium=notification, and it passing in https://travis-ci.org/cfpb/capital-framework/builds/327021795?utm_source=github_status&utm_medium=notification

jimmynotjim: Thanks.

PR #737, opened 2018-01-09
Add gulp 4

anselmbradford:
## Additions

- ~~`ajv` as a required peer-dependency.~~ weird, it originally warned me this needed to be installed, but doesn't appear to anymore. I've removed it.

## Removals

- Removes `run-sequence` and replaces it with native `gulp.series` and `gulp.parallel`.

## Changes

- Updates gulp 3.9.1 to gulp 4.0.0.
- Updates `chalk`, `gulp-autoprefixer`, `gulp-istanbul`, `gulp-less`, `gulp-markdown`, `promisify-node`, `uglifyjs-webpack-plugin`.
- Moves `build` and `default` gulp tasks inline in gulpfile to handle order of tasks definitions.

## Testing

- `npm install && gulp build && gulp test && gulp test:unit && gulp lint` (lint will have existing errors)


anselmbradford:
@jimmynotjim I decided to leave the tasks where they were and inline the build and default tasks into the gulpfile, which means less files are moving around. I think that works, since those two tasks don't add their own functionality, but just call other tasks, so having them in the gulpfile works organizationally.

scripts/gulp/copy.js
```diff
@@ -15,7 +15,11 @@ const baseManifest = JSON.parse( fs.readFileSync( './package.json', 'utf8' ) );
  * @returns {Object} An output stream from gulp.
  */
 function copyComponentsBoilerplate() {
-  return gulp.src( [ './src/' + ( parseComponentName || '*' ), '!./src/*.js', '!./src/*.less' ] )
+  return gulp.src( [
+    './src/' + ( parseComponentName || '*' ),
+    '!./src/*.js',
+    '!./src/*.less'
+  ] )
```
jimmynotjim: Should L:19-22 be indented 2 more spaces to align with L:23?

anselmbradford: I don't think so. Closing brackets/parens appear on the same indent level as where they are opened. Otherwise we'd be indenting L:29 too, which I don't think would make sense.

anselmbradford: Err I see that needs to be fixed in the lint.js

PR #751, opened 2018-02-05
Reduce H4 size on extra-small screens

Scotchester:
Per discussion and decision in https://[GHE]/CFPB/hubcap/issues/65

@nataliafitzgerald @huetingj 

## Changes

- **cf-core:** [PATCH] Reduce size of `h4` and `.h4` on extra-small screens.

## Testing

1. See https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-components-locally

## Screenshots

screen shot 2018-02-05 at 11 28 24

## Checklist

### DIFF
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -291,6 +291,14 @@ h4,
     .h6 + & {
         margin-top: unit( 30px / @font-size, em );
     }
+
+    .respond-to-max( @bp-xs-max, {
+        @h4-font-size-on-xs: 16px;
+
+        margin-bottom: unit( 10 / @h4-font-size-on-xs, em );
+        font-size: unit( @h4-font-size-on-xs / @base-font-size-px, em );
+        line-height: unit( 18 / @h4-font-size-on-xs );
+    } );
 }
 
 h5,
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -293,11 +293,11 @@ h4,
     }
 
     .respond-to-max( @bp-xs-max, {
-        @h4-font-size-on-xs: 16px;
+        @h4-font-size-on-xs: @base-font-size-px;
 
-        margin-bottom: unit( 10 / @h4-font-size-on-xs, em );
+        margin-bottom: unit( 10px / @h4-font-size-on-xs, em );
         font-size: unit( @h4-font-size-on-xs / @base-font-size-px, em );
-        line-height: unit( 18 / @h4-font-size-on-xs );
+        line-height: unit( 18px / @h4-font-size-on-xs );
     } );
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 -
 
 ### Changed
--
+- **cf-core:** [PATCH] Reduce size of `h4` and `.h4` on extra-small screens.
 
 ### Removed
 -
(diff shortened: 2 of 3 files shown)


nataliafitzgerald:
Looks good @Scotchester!

src/cf-core/src/cf-base.less
```diff
@@ -291,6 +291,14 @@ h4,
     .h6 + & {
         margin-top: unit( 30px / @font-size, em );
     }
+
+    .respond-to-max( @bp-xs-max, {
+        @h4-font-size-on-xs: 16px;
```
jimmynotjim: Presumably this should always match the body size, can we use `@base-font-size-px` here? Otherwise it's not very discoverable.

src/cf-core/src/cf-base.less
```diff
@@ -291,6 +291,14 @@ h4,
...
+
+        margin-bottom: unit( 10 / @h4-font-size-on-xs, em );
```
jimmynotjim: Should `10` have a unit type?

src/cf-core/src/cf-base.less
```diff
@@ -291,6 +291,14 @@ h4,
...
+        font-size: unit( @h4-font-size-on-xs / @base-font-size-px, em );
+        line-height: unit( 18 / @h4-font-size-on-xs );
```
jimmynotjim: Not sure there's a need for `unit()` here, would `18 / @h4-font-size-on-xs` have the same output?

src/cf-core/src/cf-base.less
Scotchester: Eh, I'm not 100% sure that's true, but it definitely should never drop _below_ the body size, so setting it to that would prevent it from doing so if, say, we ever increased the body size to 18. So yeah, I'll make that change.

src/cf-core/src/cf-base.less
Scotchester: It's apparently not necessary, but if our convention is to include one, I'll add it. I admit I didn't look elsewhere for prior art.

src/cf-core/src/cf-base.less
Scotchester: Another case where I should've looked at prior art :) I'll find out.

src/cf-core/src/cf-base.less
jimmynotjim: I was going off the helper text "displays at same size as body copy"

Scotchester: Ah, I see. In that case, I was following the form of the bigger headings' descriptions, which say "displays at same size as h<-1>". The scenario I'm pondering where it might not be true in the future is if we decided we wanted to adjust the hierarchy so that all four levels and the ghost-level for h4s on mobile were above the body size. But this will do for now.

src/cf-core/src/cf-base.less
Scotchester: It does seem that we usually include the `px` most of the time.

src/cf-core/src/cf-base.less
Scotchester: It looks like we need the `unit` to ensure any units in the component parts of a calculation get dropped, not carried through to the result. Your suggestion would result in `line-height: 1.125px;`.

jimmynotjim: Ahhh, I thought Less dropped units if only one value had them. Thanks for checking in on it.

src/cf-core/src/cf-base.less
Scotchester: Updates pushed. We could conceivably do away with the `@h4-font-size-on-xs` variable at this point and just refer to `@base-font-size-px` in the three properties, or we could leave it in place in case the scenario I described ever comes to pass. Do you have a strong preference?

PR #722, opened 2017-12-26
Remove Avenir Next Demibold web font, replace with Medium + CSS effects

cfarm:
Removes Avenir Next Demibold web font to improve performance for end users. Demibold is replaced with Avenir Next Medium + CSS effects to maintain the existing type hierarchy. For more background see https://GHE/CFGOV/platform/issues/2347#issuecomment-144964 

## Additions

-

## Removals

-

## Changes

Text elements previously using Demibold are mainly 2 types: h5-style headings, and inline bold text. Both of these were updated as part of this PR.

### H5 and elements that inherit its styling:

h5 and h6 now use Avenir Next Medium plus a CSS `text-shadow` effect to mimic the thicker weight of Demibold as compared to Medium. Additional elements inherit this CSS:

- `th` (Table heading) elements on tables (cf-core)
- Responsive stacked table headings (cf-tables)
- Slug header (cf-typography)
- Pull quote author - text-shadow color is set to match the gray text color instead of default black of heading-5 elements (cf-typography)
- Date, with text-shadow updated to gray (cf-typography)

### Inline bold:

Inline bolded text - e.g., text in paragraphs **wrapped in `<b>` or `<strong>` tags for emphasis** - was previously styled using Demibold, under the `.u-webfont-demi` class. To replace the web font but maintain the level of contrast needed between **bold text** and regular text, the Avenir Next Medium font is used, with `font-weight: bold` added to the class. The class name `u-webfont-demi` is still used to maintain consistency with our print and visual design type hierarchy. We just fake the demibold without actually making the user load that extra font file. 😎 

## Testing

Use [testing instructions](https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-components-locally) to run this locally with CF docs and confirm the above elements match screenshots for the following component docs:

- http://localhost:3000/components/cf-typography/
- http://localhost:3000/components/cf-tables/
- http://localhost:3000/components/cf-core/

## Screenshots

To make this PR easier to review I'm posting screenshots in the comments.

## Notes

-

## Todos

-

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
[...]

PR #753, opened 2018-02-08
Demibold fix for Safari

cfarm:
Builds on changes from #722 but adjusts the font-weight so that it appears the same as Demibold in all browsers, including Safari.

## Additions

- 

## Removals

-

## Changes

- `font-weight: bold` was changed to `600` after @Scotchester noticed the bold was too heavy in Safari (both desktop and Mobile Safari showed this issue). After making this change, it was tested and visually QA'ed in all of our supported browsers, and I've confirmed the change appears as expected, with the faux-Demibold now very closely matching the appearance of the web font across our supported browsers.

## Testing

You can test this on the CF docs and check out the following pages to confirm the heading-5 and inline bold elements look as expected:
- http://localhost:3000/components/cf-typography/
- http://localhost:3000/components/cf-core/

I've also tested this extensively in cfgov-refresh using npm link and Sauce labs, see screenshots below. If you test in cfgov-refresh, here's a good URL that contains a lot of the elements that use this style: http://localhost:3000/ask-cfpb/how-can-i-get-a-refund-on-a-product-or-service-i-purchased-with-my-credit-card-en-1969/

### Here's the bug in desktop Safari:

screen shot  demibold safari bug

screen shot 2018-02-08 at 4 44 56 pm

### Here it is fixed

With screenshots from a variety of browsers. Not all browsers are included here, this is just a smattering to give you the idea - it's consistent.
### Desktop Safari, Capital Framework
screen shot  demibold safari fixed

### Desktop Safari, cfgov-refresh
 The demibold elements are the date, the "Answer" heading, the inline bolded text like where it says "Note:" and the slug headers in the sidebar ("Related questions").

screen shot 2018-02-08 at 4 46 15 pm

### Mobile Safari on iPhone:
screenshot demibold mobile safari iphone fixed

### Windows Firefox:
screenshot demibold windows firefox fixed

### Windows Edge:
screenshot demibold windows edge fixed

## Notes

-

## Todos

-

## Checklist

### Browsers

- [x] Chrome
- [x] Safari
- [x] FF
- [x] IE10
- [x] IE9
- [x] IE8
- [ ] Opera
- [x] iOS
- [x] Android
- [ ] Blackberry Bold

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -120,7 +120,7 @@ textarea {
 
     margin-bottom: unit( 15px / @font-size, em );
     font-size: unit( @font-size / @base-font-size-px, em );
-    font-weight: bold;
+    font-weight: 600;
     letter-spacing: 1px;
     line-height: 1.25;
     text-shadow: 0px 0px @text-shadow;
@@ -132,7 +132,7 @@ textarea {
 
     margin-bottom: unit( 15px / @font-size, em );
     font-size: unit( @font-size / @base-font-size-px, em );
-    font-weight: bold;
+    font-weight: 600;
     letter-spacing: 1px;
     line-height: 1.25;
     text-shadow: 0px 0px @text-shadow;
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -70,6 +70,11 @@ body {
     line-height: @base-line-height;
 }
 
+strong,
+b {
+    font-weight: 600;
+}
+
 button,
 input,
 select,
--- a/src/cf-core/src/cf-base.less
+++ b/src/cf-core/src/cf-base.less
@@ -70,11 +70,6 @@ body {
     line-height: @base-line-height;
 }
 
-strong,
-b {
-    font-weight: 600;
-}
-
 button,
 input,
 select,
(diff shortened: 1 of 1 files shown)


src/cf-core/src/cf-base.less
```diff
@@ -70,6 +70,11 @@ body {
     line-height: @base-line-height;
 }
 
+strong,
+b {
+    font-weight: 600;
+}
```
Scotchester: Could you please move this below the `button, input, select, textarea` rule? I'd like to keep that rule together with the `body` rule because of the unusual nature of having to set `font-family` twice.

cfarm:
Since Firefox is the browser for only 3.5% of our users, we'll take the performance gain on this change to get it released over in cfgov-refresh ASAP. I'll start up a new branch to try out some FF fixes that can go out separately.

src/cf-core/src/cf-base.less
```diff
@@ -53,7 +53,7 @@
 .u-webfont-demi() {
     font-family: @webfont-medium, Arial, sans-serif;
     font-style: normal;
-    font-weight: bold;
+    font-weight: 600;
```
jimmynotjim: Are we sure this isn't loading the local version of `600`? Have we tested this on a Win machine?

Scotchester: Pretty sure it can't be using the local font, because we're using `@font-face` to declare what the source of the font family `'Avenir Next'` is, and we aren't including local sources there. There's a Windows Firefox (looks great) and Windows Edge (looks like it's just using Medium) screenshot up in the OP, though ☝️

Issue #771, opened 2018-02-27
Adjustments to checkbox and radio large target areas


nataliafitzgerald:
# Adjustments to checkbox and radio large target areas
Current implementation of [checkboxes and radio buttons with large target areas](https://cfpb.github.io/design-manual/page-components/form-fields.html#checkboxes-and-radio-buttons-with-large-target-areas) does not reflect the design intent as indicated in the DM specs and is not consistent with the logic being applied to other form components. 

#### Target area
- Border: None
- Background: Gray 10 (#e7e8e9)
#### Radio button (or checkbox)
- Border: 2 px, Pacific (#0072ce)
<img width="894" alt="screen shot 2018-02-27 at 6 15 03 pm" src="https://user-images.githubusercontent.com/6887128/36760767-2c364f36-1bea-11e8-84b4-d27f35ff1b7a.png">

#### Target area
- Border: None
- Background: Gray 10 (#e7e8e9)
- Outline: Dotted 1px, Pacific (#0072ce)
#### Radio button (or checkbox)
- Border: 2 px, Pacific (#0072ce)
<img width="881" alt="screen shot 2018-02-27 at 6 15 07 pm" src="https://user-images.githubusercontent.com/6887128/36760775-31015a2e-1bea-11e8-9fb4-4b1b6c29172c.png">

#### Target area
- Border: None
- Background: Pacific 20 (#d6e8fa)
#### Radio button (or checkbox)
- Border: 1 px, Gray 60 (#919395)
- Background: 14 x 14 px, Pacific (#0072ce)
<img width="973" alt="screen shot 2018-02-28 at 11 01 30 am" src="https://user-images.githubusercontent.com/6887128/36797906-307301ce-1c77-11e8-9a1b-90c9c3dacaed.png">

#### Target area
- Border: 2 px, Pacific (#0072ce)
- Background: Gray 10 (#e7e8e9)
#### Radio button (or checkbox)
- Border: 2 px, Pacific (#0072ce)
<img width="932" alt="screen shot 2018-02-27 at 6 14 31 pm" src="https://user-images.githubusercontent.com/6887128/36760749-1b9e9aca-1bea-11e8-91a2-87f5a4dec7d1.png">

#### Target area
- Border: 2 px, Pacific (#0072ce)
- Background: Gray 10 (#e7e8e9)
- Outline: Dotted 1px, Pacific (#0072ce)
- Outline offset: 3px
#### Radio button (or checkbox)
[...]

PR #774, opened 2018-03-02
Adjusts form inputs

anselmbradford:
Makes adjustments to form inputs per https://github.com/cfpb/capital-framework/issues/771

## Additions

- ~Adds variables `@form-field-border`, `@form-field-border__selected`, `@form-field-border__disabled`~.
- ~Adds variables `@form-field-border`, `@form-field-border__disabled`, and `@form-field-lg-target-border`~.
- Adds variables `@form-field-input-border`, `@form-field-input-border__disabled`, and `@form-field-input-lg-target-border`.

## Changes

- ~Changes radio button/checkbox variables to `@form-field-lg-` from `@input-lg-…` prefix~.
- Changes radio button/checkbox variables to `@form-field-input-lg-` from `@input-lg-…` prefix.
- Adds border (box-shadow) to large target input's hover state.
- Adjusts color of radio button/checkbox border color.

## Testing

1. Pull branch and follow https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-components-locally

## Screenshots

Before:
screen shot 2018-03-02 at 1 53 27 pm

After:
large-target


src/cf-forms/src/molecules/form-fields.less
```diff
@@ -21,7 +21,7 @@
                 grid-row-start: 1;
                 grid-row-end: 3;
 
-                border: 1px solid @input-border;
+                border: 1px solid @form-field-border;
```
jimmynotjim: This is the gray border for the box and circle, are we sure we want to call it `form-field-border`?

jimmynotjim: I'm not sure this is even right for the large target inputs. Form fields wrap lots of elements, including the labels paired with text inputs and select elements. Since these really are borders for the input label, not the form-field, they should probably stay as "input-#"

anselmbradford: That's okay, but we need something that is only used for the radio buttons/checkboxes in `form-fields.less`, and isn't shared across `select.less` and `text-input.less`. Otherwise it's too easy to be working on one type of input and inadvertently change a color on another. What name do you think?

anselmbradford:
New appearance per feedback from Natalia
screen shot 2018-03-02 at 5 25 24 pm

src/cf-forms/src/molecules/form-fields.less
jimmynotjim: These pseudo elements are a faux input, created because we couldn't reliably style the radio and checkbox input types (are we certain that's still the case without worrying about IE7, 8, or 9???). They should continue to use the standard input border variables because they use the same colors as a text input correct? 

For the large target labels, how about `large-target-border`, that way they're never highjacked for anything else.

nataliafitzgerald:
@anselmbradford 
Can you add a screenshot for the large target area radio buttons?

anselmbradford:
Accesslint reports can be ignored as we duplicate code in the usage doc.

anselmbradford:
@nataliafitzgerald Sorry, I realized I never got you the screenshots here.

screen shot 2018-11-06 at 10 08 15 am
screen shot 2018-11-06 at 10 08 26 am

src/cf-forms/src/molecules/form-fields.less
Scotchester: `@large-target-widgets-border`?

anselmbradford: I feel like we want to have some connection between the CSS class and the variable it uses. I named these `@form-field-*` because they are used within `m-form-field`, like how `@input-*` is used in `a-text-input` and `@select-*` is used in `.a-select`. My point above is that if we have variables stretched across different classes it makes updates brittle, and it may be an anti-pattern (pending something like ye olde visual regression testing), so if they aren't used in another class, why not name them after the class within which they're used? I changed `@input-lg-target-bg` to `@form-field-lg-target-bg` under this premise because it is used in `m-form-field`, not `a-text-input` (also, anything with large in the name should probably follow this existing format and order, so `@large` should not come first).

anselmbradford:
Rebased with canary and updated OP after screenshot and changes.

src/cf-forms/src/molecules/form-fields.less
virginiacc: Could something like `@form-field-input-border` work to add a bit more context? Otherwise, I think this is fine and shouldn't prevent merging.

anselmbradford: @virginiacc Good idea. I've updated it to that. We may revisit the naming overall for forms, but out of scope for this PR.

PR #775, opened 2018-03-02
Add cf-modals

anselmbradford:
Adds https://github.com/cfpb/capital-framework/pull/583/ to prerelease-ver-6 branch


src/cf-modals/src/Modal.js
```diff
@@ -0,0 +1,172 @@
+/* ==========================================================================
+   Modal Organism
+   ========================================================================== */
+
+const UNDEFINED = require( 'cf-atomic-component/src/utilities/config' ).UNDEFINED;
```
anselmbradford: I don't see why [this is fine](https://github.com/cfpb/capital-framework/blob/canary/src/cf-expandables/src/Expandable.js#L18), but it's not fine here.

anselmbradford:
@jimmynotjim did we figure out on Friday that the issue here in Travis is that cf-modals is not on npm yet?

jimmynotjim:
@anselmbradford No, I don't think that's the case. I've got a bit of time so I'll see if I can track this down locally.

jimmynotjim:
Figured it out, you used `atomic-component` in [package.json](https://github.com/cfpb/capital-framework/blob/canary/src/cf-expandables/package.json#L13) instead of `cf-atomic-component`. Update that and it'll build fine.

jimmynotjim:
Since this was already approved and is going into the prerelease branch I'm going to merge it. If anyone has comments we can address them in a new PR.

PR #786, opened 2018-03-16
Initial list of default reviewers in alpha order

jimmynotjim:
Lets see how this goes for PRs going forward. If it gets too noisy maybe we can divvy up sections of the project.

## Additions

- Adds GitHub codeowners file.

## Notes

- https://help.github.com/articles/about-codeowners/

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,19 @@
+# Uses the same pattern rules for gitignore files
+# https://git-scm.com/docs/gitignore#_pattern_format
+# For more info about CODEOWNERS, visit
+# https://help.github.com/articles/about-codeowners/
+
+#
+# These are the default owners for everything in the repo.
+#
+* @anselmbradford @cfarm @contolini @jimmynotjim @Scotchester @sebworks
+
+# To add owners of a specific file type or directory, prepend the line with
+# the pattern to match. The last matching pattern takes the most precedence.
+# Example:
+# *.js @js-owner
+
+#
+# These are the owners of the deployment process
+#
+/scripts/npm/prepublish/ @contolini @jimmynotjim


.github/CODEOWNERS
```diff
@@ -0,0 +1,7 @@
+# See https://help.github.com/articles/about-codeowners/
+# for more info about CODEOWNERS file
+
+# It uses the same pattern rule for gitignore file
+# https://git-scm.com/docs/gitignore#_pattern_format
+
+* @anselmbradford @cfarm @contollini @jimmynotjim @Scotchester
```
Scotchester: You've got an extra `l` in @contolini.

Scotchester: It might be nice to include a comment making it clear that this is setting the default owners, and that more specific rules could be added below that would take precedence.

contolini: If I had a nickel...

jimmynotjim: @contolini ha, maybe you should change your name to match our mental model of how it's spelled ;) I've fixed it and amended the original commit.

jimmynotjim: Good point, I've reworked the doc a bit to better explain it and added a working rule for the deployment script.

.github/CODEOWNERS
```diff
@@ -0,0 +1,18 @@
+# Uses the same pattern rule for gitignore file
```
Scotchester: Trivial: let's say either "rules" or "format", and "files".

.github/CODEOWNERS
```diff
@@ -0,0 +1,18 @@
...
+# https://git-scm.com/docs/gitignore#_pattern_format
+# For more info about CODEOWNERS, visit
+# https://help.github.com/articles/about-codeowners/.
```
Scotchester: Remove the period at the end, just so there's no ambiguity

.github/CODEOWNERS
```diff
@@ -0,0 +1,18 @@
...
+
+#
+# These are the default owners for everything in the repo.
+#
+* @anselmbradford @cfarm @contolini @jimmynotjim @Scotchester @sebworks
+
+# To add owners of a specific file type or directory, prepend the line with
+# the type to match. The last matching pattern takes the most precedence. Ex.
+# *.js                      @js-owner
```
Scotchester: Replace "type" with "pattern"

.github/CODEOWNERS
```diff
@@ -0,0 +1,18 @@
...
+
+#
+# These are the owners of the deployment process
+#
+/scripts/npm/prepublish/    @contolini @jimmynotjim
```
Scotchester: I think we should stick with a consistent one or two spaces between pattern and users. It's weird that the global rule can't match this or its line length would be too long.

jimmynotjim: Addressed all of these comments in the latest push.

.github/CODEOWNERS
```diff
@@ -0,0 +1,18 @@
...
+# For more info about CODEOWNERS, visit
+# https://help.github.com/articles/about-codeowners/
+
+#
+# These are the default owners for everything in the repo.
+#
+* @anselmbradford @cfarm @contolini @jimmynotjim @Scotchester @sebworks
+
+# To add owners of a specific file pattern or directory, prepend the line with
+# the type to match. The last matching pattern takes the most precedence. Ex.
```
Scotchester: I'd still really like to see this say "pattern" instead of "type", since it can be based on things other than file type.

jimmynotjim: Ha, I thought you meant the usage of "type" on L:11, didn't even see L:12. I thought it was odd but just went with it, realizing now it makes a lot more sense this way.

anselmbradford:
Dang we get little shields 
screen shot 2018-03-19 at 2 36 01 pm

jimmynotjim:
It's like Zelda up in here.

PR #787, opened 2018-03-22
Updated expandables to fix style bugs and follow Atomic Design principles

### LINKED ISSUE #779: Update H4 in expandables
In updating the H4 for small screens we appear to have missed expandables.  

**Related issues and links:** 
https://github.com/cfpb/capital-framework/pull/751
https://[GHE]/CFPB/hubcap/issues/65
https://cfpb.github.io/design-manual/brand-guidelines/typography.html#web-hierarchy

## Current behavior
Desktop: H4 (Avenir Next Medium 18px / 22px)
< 601px wide: H4 (Avenir Next Medium 18px / 22px)

## Expected behavior
Desktop: H4 (Avenir Next Medium 18px / 22px)
< 601px wide: H4 (Avenir Next Medium 16px / 18px)

## Steps to replicate behavior
1. https://www.consumerfinance.gov/consumer-tools/prepaid-cards/choose-the-right-card/

## Screenshots
__Current__

| Desktop | Mobile | 
| --- | --- | 
| 5 px| 10 px|

__Expected__

| Desktop | Mobile | 
| --- | --- | 
| 5 px| 10 px|

### LINKED ISSUE #780: Incorrect styling for gray expandable
The styling for our gray expandable variation is incorrect. Additionally, the gray expandable version should be in the variations section within the docs. The white background version should be the default. 

## Current behavior
- Border: No border
- Background (active): Gray 10 (#d2d3d5)
- Background (hover): Gray 20 (#e7e8e9)
- Font-size: 1em
screen shot 2018-03-08 at 12 01 34 pm
screen shot 2018-03-08 at 12 01 37 pm

## Expected behavior
- Border: Gray 40 (#b4b5b6)
- Background (active): Gray 5 (#f7f8f9)
- Font-size: 1.125em
- Background (hover): Gray 10 (#e7e8e9)
screen shot 2018-03-08 at 12 16 00 pm
screen shot 2018-03-08 at 12 15 05 pm

## Steps to replicate behavior
1. https://cfpb.github.io/capital-framework/components/cf-expandables

Related content: 
https://cfpb.github.io/design-manual/page-components/expandables.html
https://github.cfpb.gov/flapjack/Modules-V1/wiki/Expandables (DM specs will also need to be updated once we make this fix so please don't reference those specs)

jimmynotjim:
- Fixes text sizes of expandable groups (resolves #779)
- Fixes background and border colors (resolves #780)
- Removes margin outside the expandable and expandable group

## Additions

- Broke background and border modifications to their own classes.

## Removals

- Removed the margin, the component should not modify layout outside of it's bounds and the margin should be determined by the template.

## Changes

- Updated the recommended markup template to use h4

## Testing

1. Navigate to `http://localhost:3000/components/cf-expandables/#default-state`

## Screenshots

screen shot 2018-03-27 at 1 45 59 pm

screen shot 2018-03-27 at 1 46 07 pm

screen shot 2018-03-22 at 12 34 59 pm

screen shot 2018-03-22 at 12 35 06 pm

## Notes

- This should probably be considered a breaking change. I've left off the changelog for the time being. These are the ideal updates to make this component follow our existing Atomic Design principles, but I'm open to discussing non-breaking suggestions.
- Should be released with #810

## Checklist

### Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/src/cf-expandables/src/cf-expandables.less
+++ b/src/cf-expandables/src/cf-expandables.less
@@ -17,14 +17,9 @@
 // .o-expandable_link
 @expandable_link-text:          @pacific;
 
-// .o-expandable__padded
-@expandable__padded-bg:         @gray-10;
-@expandable__padded-bg-hover:   @gray-20;
-@expandable__padded-divider:    @gray-40;
-
-// .o-expandable-group
-@expandable-group-bg:           @white;
-@expandable-group-divider:      @gray-40;
+// .o-expandable modifiers
+@expandable__background:        @gray-5;
+@expandable__border:            @gray-40;
 
 // Sizing variables
 
@@ -50,7 +45,6 @@
 
 .o-expandable {
     position: relative;
-    margin-bottom: unit( 30px / @base-font-size-px, em );
 
     &_target {
         padding: 0;
@@ -63,18 +57,18 @@
             outline-offset: 1px;
         }
 
-        &__expanded .o-expandable_cue-open {
+        .o-expandable_cue-close,
+        .o-expandable_cue-open {
             display: none;
         }
 
-        &__collapsed .o-expandable_cue-close {
-            display: none;
+        &__expanded .o-expandable_cue-close {
+            display: block;
         }
 
-        .lt-ie9 & .o-expandable_cue {
-            display: none;
+        &__collapsed .o-expandable_cue-open {
+            display: block;
         }
-
     }
 
     &_content {
@@ -101,6 +95,9 @@
     //
 
     &_label {
+        // Remove default h4 margin style
+        margin-bottom: 0;
+
         color: @expandable_label-text;
         font-weight: 500;
     }
@@ -145,38 +142,41 @@
     //
 
     &__padded {
-        margin-bottom: unit( 30px / @base-font-size-px, em );
-        background: @expandable__padded-bg;
-
-        &:hover,
-        &.o-expandable__expanded {
-            background: @expandable__padded-bg-hover;
-        }
-
         .o-expandable_header {
             padding: unit( 10px / @base-font-size-px, em )
                      unit( 15px / @base-font-size-px, em );
         }
 
         .o-expandable_content {
-            // Margin collapsing trick for more consistent bottom padding.
-            margin: 0
-                    unit( 15px / @base-font-size-px, em )
-                    unit( 22px / @base-font-size-px, em );
+            padding: 0 unit( 15px / @base-font-size-px, em );
 
             // The divider between _header and _content.
             &:before {
                 content: '';
                 display: block;
-                border-bottom: 1px solid @expandable__padded-divider;
-                margin-bottom: unit( 15px / @base-font-size-px, em );
+                border-top: 1px solid @expandable__border;
+                padding-top: unit( 15px / @base-font-size-px, em );
             }
 
             &:after {
-                margin-bottom: unit( 15px / @base-font-size-px, em );
+                padding-bottom: unit( 15px / @base-font-size-px, em );
                 width: 100%;
             }
         }
+    }
+
+    //
+    // Expandable with a background color modifier
+    //
+    &__background {
+        background: @expandable__background;
+    }
+
+    //
+    // Expandable with a border modifier
+    //
+    &__border {
+        border: 1px solid @expandable__border;
     }
 
     //
@@ -184,27 +184,12 @@
     //
 
     &-group {
-        margin-bottom: unit( 30px / @base-font-size-px, em );
-
-        .o-expandable {
-            border-bottom: 1px solid @expandable-group-divider;
-            margin-bottom: 0;
-            background: @expandable-group-bg;
+        .o-expandable__padded {
+            border-bottom: 1px solid @expandable__border;
 
             &:first-child {
-                border-top: 1px solid @expandable-group-divider;
+                border-top: 1px solid @expandable__border;
             }
         }
-
-        .o-expandable_label {
-            .heading-4();
-            margin-bottom: 0;
-        }
-
-        .o-expandable_content {
-            margin-bottom: 0;
-        }
-
     }
-
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 -
 
 ### Changed
--
+- **cf-expandables:** [MAJOR] Update the expandables to fix style bugs and follow Atomic Design principles
 
 ### Removed
 -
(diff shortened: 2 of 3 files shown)


src/cf-expandables/src/cf-expandables.less
```diff
@@ -17,14 +17,9 @@
...
-// .o-expandable__padded
-@expandable__padded-bg:         @gray-10;
-@expandable__padded-bg-hover:   @gray-20;
-@expandable__padded-divider:    @gray-40;
-
-// .o-expandable-group
-@expandable-group-bg:           @white;
-@expandable-group-divider:      @gray-40;
+// .o-expandable modifiers
+@expandable__background:        @gray-5;
```
sebworks: Shouldn't this be `@expandable__gray5-background`.  This would allow you to have different color backgrounds?

jimmynotjim: The background should not be any other color according to the spec.

src/cf-expandables/src/cf-expandables.less
```diff
@@ -50,7 +45,7 @@
 
 .o-expandable {
     position: relative;
-    margin-bottom: unit( 30px / @base-font-size-px, em );
+    //margin-bottom: unit( 30px / @base-font-size-px, em );
```
anselmbradford: Can we delete this?

anselmbradford: Here and the other commented out margin-bottom bits.

src/cf-expandables/src/cf-expandables.less
```diff
@@ -145,66 +140,63 @@
...
+            //margin-bottom: 0;
 
             &:first-child {
-                border-top: 1px solid @expandable-group-divider;
+                border-top: 1px solid @expandable__border;
             }
         }
 
         .o-expandable_label {
-            .heading-4();
```
sebworks: Why remove this and use the `h4` class instead?

jimmynotjim: Because this mixin doesn't account for the change in font size on smaller screens. See issue #779

src/cf-expandables/src/cf-expandables.less
jimmynotjim: I left them as part of the discussion about breaking changes, but my intention is to delete them.

sebworks:
@jimmynotjim, would it be possible to include only what we needed from cf-core.less,  cf-buttons.less, and cf-icons.less?

src/cf-expandables/src/cf-expandables.less
sebworks: ok, I think it complicates the module because it gives the impression that the heading size is flexible.

jimmynotjim: Weirdly, it is. Solo expandables don't have a modified font size or weight, it appears that's only for expandable groups. I'm not sure of the reason, but maybe @nataliafitzgerald can fill us in.

jimmynotjim:
>  would it be possible to include only what we needed from cf-core.less, cf-buttons.less, and cf-icons.less?

I've always wanted to figure out how to do this globally, because it would remove a lot of the reasoning behind using individually packaged and distributed modules instead of one installable for everything and make it a lot easier to get started using or contributing to CF. I'm not sure it's possible with our current toolchain due to the nature of Less. It may be worth exploring as we continue the discussions around combining CF and the Design Manual.

src/cf-expandables/src/cf-expandables.less
jimmynotjim: I also should have mentioned that because we're not modifying the h4 mixin in any way that using the class in the recommended markup pattern reduces the duplication of output which is a big culprit to our css file size woes.

jimmynotjim:
> would it be possible to include only what we needed from cf-core.less, cf-buttons.less, and cf-icons.less?

I just had an `aha` moment on this. What if we pulled the component variables out of the main component file? If all you need is `cf-buttons/src/atoms/buttons.less`, you could import the variables file, then the atomic/buttons.less file, and ignore the rest. The main `cf-buttons/src/cf-buttons.less` would then add an import for the variables instead of defining those variables itself. I don't know how well Less will handle one module importing everything and another importing just some of the things, but we already import cf-core in all the other modules so it _seems_ like it figures it all out and only includes the cf-core files once. At this point I should probably move this discussion to it's own issue huh?

jimmynotjim:
Let's move the `import` discussion over to #788

src/cf-expandables/src/cf-expandables.less
sebworks: yeah, it is part of my confusion surrounding when to use margin utilities within atomic components. If an interior DOM element needs some margin ( same across all breakpoints ), should we use the utility margin classes or just handle it with CSS?

jimmynotjim: I feel like utility classes are a bit different, they do one thing, and one thing well, but that thing could also usually be packaged with the other things happening in that element to help describe the situation. IE:

```less
.m-thing {
  padding: 15px;
  background-color: @gray-5;

  a-atom {
    margin-bottom: 30px;
  }
}
```
In this example the margin is the only style that could be offloaded to the markup, providing minimal bloat reduction. On the other hand, keeping it in the Less provides the next developer a lot more context of how the pattern works as a whole.

For something complex like h# classes that include font styles, margin, letter-spacing, and breakpoint changes, putting them into the markup removes far more bloat and the font-styles don't have any affect on the rest of the pattern so there isn't a loss of context by removing them from the Less.

jimmynotjim:
> Should we change our example markup to reflect that?
Hmm, this makes sense to me. I'll make an update.

It looks like modals might be a while yet. Since this would require a breaking change, is it worth holding off merging till either modals or another breaking change (I know icons is coming as well) are ready?

Scotchester:
I think we should go with this and icons.

jimmynotjim:
@Scotchester I've updated the examples to use `h3`s since it seems that's the typical heading hierarchy when expandables are used within content areas. I'm not sure how we should handle this on the wagtail admin side? We may need to add a heading level selector to allow editors to set the correct level for each individual piece of content.

Scotchester:
Yeah, we could add a heading level selector to `ExpandableGroup`, conceivably. Individual expandables within a group shouldn't need to have different heading levels. The top and bottom padding in the expandable header looks a little tight. Is that to spec?

If that's good, you're still approved; merge when ready!

jimmynotjim:
Thanks, I'm going to get the new alert roles working in the OaH explorer and then we should have removed all the old font icons. At that point I'll feel comfortable releasing the next major update here and not worrying about breaking things.

PR #797, opened 2018-04-18
Add an animated version of the update icon

jimmynotjim:
When we released the svg icons we skipped the animated icons for the time
being. Turns out the update icon is realistically the only one that
needs to be animated. Duplicating the static icon with a modifier class
and then targeting that in the Less is the simplest solution.

## Additions

- `cf-icon-updating` SVGs
- `cf-icon-svg__updating` styles

## Testing

1. Navigate to `http://localhost:3000/components/cf-icons/#status-icons`
1. Modify the class of the existing updating icon to include `cf-icon-svg__updating` (there's a weird jekyll bug where if I added the new icons on their own row it would render each and the name as the first column of multiple rows. Will investigate a fix outside of this PR).

## Checklist

### Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Internet Explorer 8, 9, 10, and 11 (where animation is supported otherwise it's static)
- [x] Edge
- [x] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [x] Is useable without JS


src/cf-icons/src/cf-icons.less
```diff
@@ -38,6 +38,20 @@
...
+        transform-origin: 50% 50%;
+    }
+}
+
+@keyframes updating-animation {
+    0% {
+        transform: rotate( 0deg );
+    }
+    100% {
+        transform: rotate( 359deg );
```
wpears: I feel like this should be `transform: rotate( 360deg );`, though I haven't seen a case made either way. Conceptually, only going to 359 would mean that there'd be a 1 degree jump at the start of the loop (from 359 to 0), but I know there are certain animation frame hacks that require dropping a degree to make the end result seem smoother, but I dunno if that is true in this case.

sebworks:
Is it worth using`will-change`? I know it isn't supported by all the browsers though.

src/cf-icons/src/cf-icons.less
jimmynotjim: No, 360 and 0 are the same position, so if you do 360 there's a pause at the end rather than smooth animation.

jimmynotjim:
> Is it worth using `will-change`? I know it isn't supported by all the browsers though.

There's no reason to use `will-change` on the icon, because the icon is always animated. The entire `svg` is replaced if the item is no longer being updated. Notification container should probably get `will-change` though, because it's typically changing icons and text.

PR #810, opened 2018-07-09
HOLD: Remove the deprecated font-icon system

jimmynotjim:
The font-icons were deprecated when the SVG icon system was released.
Now that most projects have been updated it's time to remove the old
icon font and related styles and documentation.

## Removals

-  Removed the deprecated font-icon system from cf-icons

## Testing

1. Navigate to `http://localhost:3000/components/cf-icons`
1. Confirm the icon font documentation is gone.

## Notes

- Should be released along with #787 

## Todos

- Double check cfgov-refresh has not had font icons added back into the project since it was updated for CFv5

## Checklist

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Internet Explorer 8, 9, 10, and 11
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


jimmynotjim:
@anselmbradford thanks for catching that, completely forgot about it. Updated with the change.

jimmynotjim:
Holding to join with #787 for a single major release of CF.

jimmynotjim:
Turns out there are many icons added via Wagtail to cfgov-refresh, so this PR will need to stay on hold until they can either be converted to SVG icons or removed.

Scotchester:
@anselmbradford We need to search Wagtail content fields for places where icon-font icons have been manually added via `<span class="cf-icon-whatever"></span>` tags and replace those with SVGs.

In cases where we're not concerned about a nontechnical person's ability to edit the content regularly, we can probably just dump in the raw SVG code. If we find there are other cases where that would be extremely undesirable, I would advocate for revisiting my (potentially controversial ::cough @chosak cough::) shortcode idea – putting in something like [cf-icon-whatever] that would be replaced on the back end with the SVG code.

chosak:
> We need to search Wagtail content fields 

Don't we also need to update any other places ([example](https://github.com/cfpb/retirement/blob/57dc426ec195d69aedc752a0c20c0523d14ec2bd/retirement_api/templates/retirement_api/about.html#L27)) where we hardcode HTML (or, in retirement's case, even [database fixtures](https://github.com/cfpb/retirement/blob/a2a9a590a0ae2ac5bf931f89129b136578cc7ae8/retirement_api/fixtures/retiredata.json#L382) 😢) with `cf-icon`?

> we can probably just dump in the raw SVG code

We should brainstorm this a little bit -- don't we want to avoid duplicating raw SVG content all over the place? The shortcode idea could work (and would be nicer/easier in Wagtail 2 where we can more easily implement custom editor tools), but a more general Django templatetag library for `{{ svg_icon }}` could also be nice.

Scotchester:
> Don't we also need to update any other places where we hardcode HTML (or, in retirement's case, even database fixtures 😢) 

Yes, also true. I think there are none left in cfgov-refresh, but yeah, we should check all satellite apps.

Scotchester:
> We should brainstorm this a little bit -- don't we want to avoid duplicating raw SVG content all over the place?

Yes, absolutely! Also, if we dumped in raw SVG, they would have to be updated manually if we ever update an icon design. That's why I like the shortcode :) 

> The shortcode idea could work (and would be nicer/easier in Wagtail 2 where we can more easily implement custom editor tools), but a more general Django templatetag library for {{ svg_icon }} could also be nice.

I don't follow you here.

chosak:
@Scotchester I'm sorry if I wasn't clear; I wrote too quickly.

What I meant to say was that a Django template tag library (like what we have implemented as `svg_icon` in cfgov-refresh) would be a useful tool for projects to use to insert Capital Framework SVGs into pages. If this were implemented as a pip-installable Python package (say, django-capital-framework), it'd let satellite apps include SVG icons in templates more easily, without needing to duplicate the SVG content.

This could be totally independent of a "shortcode" implementation in Wagtail RichTextFields, or that implementation could somehow be built on top of the Django template tag logic.

Scotchester:
Gotcha :+1:

Scotchester:
@jimmynotjim How big would that one SVG file be, though?

jimmynotjim:
You could break it up into a few SVGs, probably by the groups they're already in, but maybe by usage makes more sense. Then you're only loading one or two per page (typically) and once they're cached they won't be loaded again.

anselmbradford:
Would it be so bad to incrementally start converting the icons over to raw svgs till a more DRY solution is determined? It seems like they could be found and replaced with a short code of some sort fairly easily in the future and it would allow [significant cleanup to happen](https://github.com/cfpb/capital-framework/blob/31c2f72a836251643186fb69dc77336c7b3c212b/src/cf-icons/src/cf-icons.less#L57-L58) here in CF.

anselmbradford:
Bump to this 👊

jimmynotjim:
> it would allow significant cleanup to happen here in CF.

Yeah it would!

Screen Shot 2019-03-18 at 8 56 20 AM

I like this idea but I'm also not the one to clean up the raw svgs :D

anselmbradford:
Closing in favor of https://github.com/cfpb/capital-framework/pull/906, which targets the `master` branch (sorry Jimmy). That branch also adds a fix for the updating icon.

jimmynotjim:
😢

PR #816, opened 2018-07-10
Fixes some linter errors and warning

anselmbradford:
## Changes

- Ignore `max-lines-per-function` in tests.
- Fix some linter errors and warnings.
- Add some jsdocs.
- Remove unused references.

## Testing

1. Run `gulp lint` on `canary`.
2. Run `gulp lint` in this branch and see that some issues are fixed.

### DIFF
--- a/scripts/npm/prepublish/index.js
+++ b/scripts/npm/prepublish/index.js
@@ -8,6 +8,7 @@ const util = require( './lib' );
 const isTravis = require( 'is-travis' );
 const componentsDir = path.join( __dirname, '..', '..', '..', 'src' );
 let componentsToPublish = [];
+let UNDEFINED;
 
 // Check git's status.
 util.getGitStatus( './' )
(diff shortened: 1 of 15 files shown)


scripts/npm/prepublish/index.js
```diff
@@ -157,12 +158,12 @@ function compareVersionNumber( component ) {
...
     }
   } ).catch( function( err ) {
-    if ( /returned 404/.test( err ) ) {
+    if ( ( /returned 404/ ).test( err ) ) {
       util.printLn.success( component + ': ' + localVersion, true );
       return {
         name:       component,
         newVersion: localVersion,
-        oldVersion: undefined
+        oldVersion: UNDEFINED
```
wpears: Rather than specifically setting undefined, can we either leave off the key or set it to null?

anselmbradford: @contolini What do you think this should be? I'll merge this PR, but we can do another PR for this.

PR #821, opened 2018-07-17
Replaces cf-tables qunit test with jest test

anselmbradford:
Replaces cf-tables qunit test with a jest test.

## Additions

- Jest Table-spec test for cf-tables.

## Removals

- Qunit Table-spec test for cf-tables.
- Deletion of modifier.initialize function from AtomicComponent.

## Testing

1. Delete `delete modifier.initialize;` from the AtomicComponent in cf-tables node_modules.
1. `gulp test:unit` should pass.
 
## Notes

- For some reason `gulp test:unit --specs=src/cf-tables/src/Table-spec.js` is failing on an open promise, but this is an existing issue.
- There are clean up changes here I'll spin out into their own PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 -
 
 ### Changed
+- **capital-framework:** [PATCH]
+  Updates `babel-jest`, `jest`, `jest-cli`, `webpack`, and `webpack-stream`.
 - **cf-atomic-component:** [PATCH] Remove "use strict" from modules.
 
 ### Removed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "capital-framework",
-  "version": "5.2.0",
+  "version": "5.3.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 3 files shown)


src/cf-atomic-component/src/components/AtomicComponent.js
```diff
@@ -64,7 +62,6 @@ assign( AtomicComponent.prototype, Events, classList, {
       if ( classList.contains( this.element, modifier.ui.base ) ) {
         if ( modifier.initialize ) {
           this.initializers.push( modifier.initialize );
-          delete modifier.initialize;
```
anselmbradford: Something in the initialization of AtomicComponent makes the state leak. Without removing this line the tests will not pass, because they will share the state of `modifier.initialize` being deleted. To test, ensure this line is present in the AtomicComponent used by cf-tables and see that the tests fail. Then `xit` all but the last test in `Table-spec.js` and see that they pass. Enable the first test (so only the first and last are running) and see that they fail again. Somehow the state is carried between tests by `Table.init()`, which is bad news. I believe we can delete this line as the tests still pass without it, but I'm not 100% certain as this appears to be a means to indicate initialization has already happened on the modifiers (instead of say, placing a flag in the DOM). @mistergone Any insight on this?

Issue #656, opened 2017-10-17
Remove qunit


anselmbradford:
We only have jquery in this project for qunit. Let's shift to mocha and chuck qunit to the curb. Initial mocha unit tests added in https://github.com/cfpb/capital-framework/pull/655

## Current behavior

- Project uses qunit.

## Expected behavior

- Project should use mocha like in cfgov-refresh.

anselmbradford:
Project actually has switched to Jest, not mocha.

PR #827, opened 2018-07-18
Remove QUnit

anselmbradford:
Fixes https://github.com/cfpb/capital-framework/issues/656

## Additions

- Adds initial Expandables jest test spec.

## Removals

- Removes QUnit.
- Removes unnused `testID` variable from Table-spec.

## Testing

1. `gulp test` should pass and have >90% coverage for cf-expandables and cf-tables.


anselmbradford:
Note that the QUnit tests acted as acceptance tests (e.g. they had HTML mocks that included the CSS and therefore included the animations, which jest unit tests do not included), therefore the tests have been reduced in number. We should add acceptance tests for the CF docs to cover this gap.

PR #840, opened 2018-07-25
Upgrade cf-atomic-component, cf-expandables, and cf-tables

### LINKED ISSUE #536: Expandable accordion doesn't expand in some cases
Bug in the accordion where it's possible to click the expandables and not have them expand. cc @mistergone as maybe he knows what's up!

## Current behavior

- Clicking the expandables repeatedly doesn't consistently expand the expandable.
accordion-bug

## Expected behavior

- Repeated clicks of the expandable does not allow the user to leave an unexpanded expandable with "hide" showing for the label.

## Steps to replicate behavior

1. Visit https://cfpb.github.io/capital-framework/components/cf-expandables/
2. Scroll to accordion example.
3. Click the first expandable repeatedly.

### LINKED ISSUE #669: Make ExpandableTransition immutable

jimmynotjim:
_This is a multi-part PR, I can try and break it apart, but it's probably best as one large MAJOR release. All of this work was discovered while attempting to update and test the expandable transitions in atomic-component._

__Core issues resolved by this PR__
1. cf-expandables and cf-tables have been self-initializing since CFv4. This goes against our core principles that components should be initialized when and where we use them. 
1. ExpandableTransition is only used by cf-expandables and should live alongside it.
1. ExpandableTransition was relying on upstream code from Expandables.js instead of being self-contained.
1. ExpandableTransition didn't include the expand/collapse begin and end events that cfgov relies on.
1. The accordion has had some interaction bugs stopping us from deploying cf-expandables to cfgov-refresh.

## Additions

- Adds expand/collapse dispatch events
- Adds unit tests for ExpandableTransition

## Removals

- Removes init code from cf-expandables and cf-tables
- Removes custom classes from Expandable
- Removes custom classes passed into ExpandableTransition

## Changes

- Simplifies open-by-default condition in ExpandableTransition
- Updates Expandable unit tests to only test Expandable code
- Updates Expandable to fix accordion feature and includes robust tests

## Testing

1. Run `gulp test:unit` everything should pass

## Notes

- Fixes #669
- Fixes #536

## Checklist

## Testing checklist


jimmynotjim:
Travis can't npm link the `cf-atomic-component` package. It's not worth testing ExpandableTransition in `cf-atomic-component` when it's going to be moved, but if we want to move forward with https://github.com/cfpb/cfgov-refresh/pull/4365 we need these changes. What say everyone, do we continue without Travis, then send out a release once it's merged so that the correct versions are working everywhere?

src/cf-atomic-component/src/utilities/transition/ExpandableTransition.js
```diff
@@ -57,6 +55,13 @@ function ExpandableTransition( element, classes ) {
    */
   function _transitionComplete() {
     this.trigger( BaseTransition.END_EVENT, { target: this } );
```
anselmbradford: I think the lifecycle should be `transitionBegin` > `expandBegin` > `expandEnd` > `transitionEnd` …and… `transitionBegin` > `collapseBegin` > `collapseEnd` > `transitionEnd`. So this should go at the bottom of `_transitionComplete`. 

Making the above moot, I'm thinking the `transitionBegin` and `transitionEnd` events don't need to be dispatched from ExpandableTransition at all, since expand and collapse are the expandable-equivalent events. From the outside, when dealing with an ExpandableTransition it should not matter that it's using a BaseTransition under the hood. By exposing the BaseTransition events, it means that if they were ever changed in BaseTransition, it is no longer guaranteed that something depending on them through ExpandableTransition wouldn't break.

src/cf-atomic-component/src/utilities/transition/ExpandableTransition.js
```diff
@@ -57,6 +55,13 @@ function ExpandableTransition( element, classes ) {
...
+
+    if ( contains( element, classObject.EXPANDED ) ) {
```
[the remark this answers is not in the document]
anselmbradford: Does this take care of it?

```js

if ( contains( element, classObject.EXPANDED ) ) {
   this.trigger( 'expandEnd', { target: this } );

   if ( element.scrollHeight > previousHeight ) {
     element.style.maxHeight = element.scrollHeight + 'px';
   }

} else if ( contains( element, classObject.COLLAPSED ) ) {
   this.trigger( 'collapseEnd', { target: this } );
}

```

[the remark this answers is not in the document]
jimmynotjim: @anselmbradford I shouldn't have assumed that would be equally complex, cause yes, it fixes it 😄

src/cf-atomic-component/src/utilities/transition/ExpandableTransition.js
jimmynotjim: I agree with the second point, updated.

test/unit-test/src/cf-expandables/src/Expandable-spec.js
```diff
@@ -1,10 +1,10 @@
 import { simulateEvent } from '../../../../util/simulate-event';
-const Expandable = require(
+const cfExpandables = require(
   '../../../../../src/cf-expandables/src/Expandable'
```
Scotchester: Can we make these paths not relative, as well?

test/unit-test/src/cf-expandables/src/ExpandableTransition-spec.js
```diff
@@ -0,0 +1,145 @@
+import { simulateEvent } from '../../../../util/simulate-event';
+const ExpandableTransition = require(
+  '../../../../../src/cf-expandables/src/ExpandableTransition'
```
Scotchester: Can we make these paths not relative, as well?

test/unit-test/src/cf-tables/src/Table-spec.js
```diff
@@ -1,5 +1,5 @@
 import { simulateEvent } from '../../../../util/simulate-event';
-const Table = require( '../../../../../src/cf-tables/src/Table' );
+const cfTables = require( '../../../../../src/cf-tables/src/Table' );
```
Scotchester: Can we make these paths not relative, as well?

src/cf-expandables/src/Expandable.js
```diff
@@ -12,11 +13,10 @@ const removeClass = domClassList.removeClass;
 const closest = require(
   'cf-atomic-component/src/utilities/dom-closest'
 ).closest;
-const ExpandableTransition = require(
-  'cf-atomic-component/src/utilities/transition/ExpandableTransition'
-);
+const fnBind = require( 'cf-atomic-component/src/utilities/function-bind' ).bind;
```
anselmbradford: This is fine for this PR, but I think we should probably deprecate the `function-bind` module since it's just needed for IE8.

src/cf-expandables/src/Expandable.js
```diff
@@ -29,80 +29,81 @@ const Expandable = Organism.extend( {
...
- */
-function onToggleAccordion() {
-  Events.trigger( 'CFAccordionClose' );
-  this.activeAccordion = true;
+  if ( this.isAccordionGroup ) {
+    if ( this.activeAccordion ) {
+      console.log( 'active accordion' );
+      this.activeAccordion = false;
+    } else {
+      console.log( 'inactive accordion' );
```
anselmbradford: Remove console.log

src/cf-expandables/src/ExpandableTransition.js
```diff
@@ -41,25 +40,28 @@ function ExpandableTransition( element, classes ) {
...
     } else {
-      previousHeight = element.scrollHeight;
-      _baseTransition.applyClass( classObject.COLLAPSED );
+      this.collapse();
     }
 
     return this;
   }
 
+  /* istanbul ignore next */
```
anselmbradford: Why's this needed?

src/cf-expandables/src/ExpandableTransition.js
```diff
@@ -94,15 +98,20 @@ function ExpandableTransition( element, classes ) {
...
-    _baseTransition.applyClass( classObject.EXPANDED );
+    _baseTransition.applyClass( CLASSES.EXPANDED );
 
     return this;
   }
 
   // Attach public events.
   this.addEventListener = Events.on;
-  this.trigger = Events.trigger;
+  this.dispatchEvent = Events.trigger;
```
anselmbradford: 👍

src/cf-expandables/src/ExpandableTransition.js
jimmynotjim: Because it's impossible to test the transitionEnd in a testing env.

src/cf-expandables/src/ExpandableTransition.js
```diff
@@ -117,11 +126,11 @@ function ExpandableTransition( element, classes ) {
...
 
+  // Public static properties.
+  this.CLASSES = CLASSES;
+
   return this;
 }
 /* eslint-enable max-lines-per-function */
 
-// Public static properties.
-ExpandableTransition.CLASSES = CLASSES;
```
anselmbradford: Why delete this in favor of `this.CLASSES = CLASSES`? It seems like `ExpandableTransition.CLASSES = CLASSES` would be preferable since it adds it to the class, not to instances of the class.

src/cf-expandables/src/Expandable.js
jimmynotjim: I found it was necessary in order to pass `this` to `_accordionActivatedHandler` without making it a public method following the example from ExpandableTransition

test/unit-test/src/cf-tables/src/Table-spec.js
```diff
@@ -112,7 +112,7 @@ describe( 'Table', () => {
 
   beforeEach( () => {
     document.body.innerHTML = HTML_SNIPPET;
-    Table.init();
+    cfTables.init();
```
anselmbradford: Seems this should be left as `Table`, unless we're updating the filename.

src/cf-expandables/src/ExpandableTransition.js
jimmynotjim: To pass the individual instance's classes on to Expandable here https://github.com/cfpb/capital-framework/pull/840/files#diff-6a2d6a4799b49a1f411abd3797d2bc18R58

test/unit-test/src/cf-tables/src/Table-spec.js
jimmynotjim: I wasn't sure what to do here, it makes sense in places it's used to name these variables `cfExpandables` and `cfTables` so it's clear we're using the real versions and not one of the other existing implementations (see the capital-framework.js file). I figured it's important to keep things consistent for whomever comes to these files later, cause consistency has been on of our biggest issues when onboarding, even with existing devs unfamiliar to a piece of the codebase.

src/cf-expandables/src/Expandable.js
jimmynotjim: Ah, sorry, I didn't look at what `fnBind` was doing under the hood. I'll see if I can include it here so we're not sending out another major release of atomic component

test/unit-test/src/cf-tables/src/Table-spec.js
jimmynotjim: Oh yeah, I guess that's true. I'll update here and elsewhere.

src/cf-expandables/src/ExpandableTransition.js
jimmynotjim: Ohhhh, duh, thank you.

PR #846, opened 2018-07-27
Remove unneeded hidden files

anselmbradford:
Bower is deprecated, so we shouldn't need its files. `.npmignore` uses `.gitignore` when present, so that shouldn't be needed.

## Removals

- `.bowerrc` and `.npmignore`.

## Changes

- Removes bower artifacts in .gitignore.
- Removes some django and python artifacts from .gitignore.


anselmbradford:
Or… we could remove the link to the compiled source in our docs 😈

PR #852, opened 2018-08-01
Add breakpoint JS and move Less breakpoint vars

anselmbradford:
Address this TODO https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/config/breakpoints-config.js#L1 so that JS and Less breakpoint values are stored near each other.

## Additions

- Add JavaScript breakpoint variables.

## Changes

- Move less breakpoint variables to their own file.

## Testing

1. `npm run build` should pass.

## Todo
- Once this is merged and released. References to breakpoints-config.js in cfgov-refresh should be updated to refer to cf-vars-breakpoints.js instead.

### DIFF
--- a/src/cf-core/src/cf-vars-breakpoints.js
+++ b/src/cf-core/src/cf-vars-breakpoints.js
@@ -1,0 +1,29 @@
+/* ==========================================================================
+   Capital Framework
+   Breakpoint JavaScript variables.
+   Matched to values for use in JavaScript matched
+   to CSS breakpoints in cf-vars-breakpoints.less
+   All values are pixel based.
+   ========================================================================== */
+
+module.exports = {
+  bpXS: {
+    min: 0,
+    max: 600
+  },
+  bpSM: {
+    min: 601,
+    max: 900
+  },
+  bpMED: {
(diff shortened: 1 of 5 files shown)


wpears:
I'm fine with this approach, but it seems to me that the intent of that comment is to use something like https://www.npmjs.com/package/less-vars-to-js to only have the vars set to a value in one place across js/less. I think these things change rarely enough and the comment linking the two files is obvious enough that the low-tech way of just including the variables separately in js/less is fine.

src/cf-core/src/cf-vars-breakpoints.js
```diff
@@ -0,0 +1,29 @@
+/* ==========================================================================
+   Capital Framework
+   Breakpoint JavaScript variables.
+   Matched to values for use in JavaScript matched
+   to CSS breakpoints in cf-vars-breakpoints.less
```
Scotchester: This sentence is kinda confusing.

anselmbradford: Okay I adjusted it!

PR #857, opened 2018-08-20
Mock child process methods in script unit tests

contolini:
The strange cf-core phantom publishing was caused by the [publish-spec.js](https://github.com/cfpb/capital-framework/blob/d264c1fb245556847d6b116780d96a86a8bd4c57/test/unit-test/scripts/npm/prepublish/lib/publish-spec.js#L17-L24)
file executing the publishing script in Travis as part of the normal
`npm test` command.

Jest spies are unconventional in that they also call the relevant
function, see link below.

> Note: By default, jest.spyOn also calls the spied method. This is different behavior from most other test libraries. If you want to overwrite the original function, you can use jest.spyOn(object, methodName).mockImplementation(() => customImplementation) or object[methodName] = jest.fn(() => customImplementation);

https://jestjs.io/docs/en/jest-object#jestspyonobject-methodname


anselmbradford:
Whoooaaa thanks for figuring that out!

PR #861, opened 2018-09-06
Replace scrollHeight with children height in ExpandableTransition.expand

Scotchester:
The purpose of this is to enable expandable heights to be recalculated while open to be _shorter_. Using the previous `scrollHeight` method, open expandables can call `expand` to _grow_ if their contents have changed, but if their contents have shrunk, `scrollHeight` still returns the rendered height from when it was at its maximum.

This is necessary for the correct height adjustments as the multiselect filters are used within a Filterable List Control expandable in cfgov-refresh (e.g., on the [Blog](https://www.consumerfinance.gov/about-us/blog/)).

## Changes

- **cf-expandables:** [PATCH] Replace `scrollHeight` with children height in `ExpandableTransition.expand`

## Testing

- To test that this doesn't negatively impact basic expandables, follow the [local testing instructions](https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-in-the-documentation-site) and look at the Expandables page in the docs site.
- `gulp test`

## Todos

- Open the PR in cfgov-refresh to test this on Filterable List Controls.

## Checklist

### Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] iOS Safari
- [x] Chrome for Android

### DIFF
--- a/src/cf-expandables/src/ExpandableTransition.js
+++ b/src/cf-expandables/src/ExpandableTransition.js
@@ -98,8 +98,14 @@ function ExpandableTransition( element ) {
   function expand() {
     this.dispatchEvent( 'expandBegin', { target: this } );
 
-    if ( !previousHeight || element.scrollHeight > previousHeight ) {
-      previousHeight = element.scrollHeight;
+    let childrenHeight = 0;
+    for ( const child of element.children ) {
+      childrenHeight += child.scrollHeight;
+    }
+
+    if ( !previousHeight || childrenHeight !== previousHeight ) {
+      // Magic number of 30 accounts for vertical padding
+      previousHeight = childrenHeight + 30;
     }
 
     element.style.maxHeight = previousHeight + 'px';
--- a/src/cf-expandables/src/ExpandableTransition.js
+++ b/src/cf-expandables/src/ExpandableTransition.js
@@ -99,9 +99,10 @@ function ExpandableTransition( element ) {
     this.dispatchEvent( 'expandBegin', { target: this } );
 
     let childrenHeight = 0;
-    for ( const child of element.children ) {
+
+    Array.prototype.forEach.call( element.children, function( child ) {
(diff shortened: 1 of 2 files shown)


Scotchester:
Question for the group:

Tests still pass, but I can't tell if I need to add a test for the change in functionality or not.

src/cf-expandables/src/ExpandableTransition.js
```diff
@@ -98,8 +98,14 @@ function ExpandableTransition( element ) {
   function expand() {
     this.dispatchEvent( 'expandBegin', { target: this } );
 
-    if ( !previousHeight || element.scrollHeight > previousHeight ) {
-      previousHeight = element.scrollHeight;
+    let childrenHeight = 0;
+    for ( const child of element.children ) {
```
anselmbradford: No IE support for `for … of`. Do we care? It may also be picked up by babel.

anselmbradford:
> Tests still pass, but I can't tell if I need to add a test for the change in functionality or not.

For animated features that coordinate with CSS we need acceptance tests, which we don't have in CF.
We could move the code you added to a separate public method (`recalculateHeight()` or something), and test that it returns expected heights with different html snippets of various heights. However, that would appear to be making a public method only for the sake of the tests, which doesn't seem right. I'd only go that way if there's a use-case for a `recalculateHeight()` method outside of the tests, otherwise acceptance tests would be the way to go.

src/cf-expandables/src/ExpandableTransition.js
Scotchester: Ah, right. I'll change it to `Array.prototype.forEach.call`.

Scotchester:
Browser/acceptance tests would be good, but as we have none currently set up for Capital Framework, I'll defer that to later.

PR #874, opened 2018-12-04
Remove classlist-polyfill,dom-class-list module, and jquery dependencies

anselmbradford:
## Removals

- `classlist-polyfill` dependencies to drop IE9 JS support.
- `dom-class-list` module to drop IE9 JS support.
- Unused `jquery` dependency.

## Testing

1. `npm run build` should pass.
2. `gulp test` should pass.
3. Set up the docs page and open in IE10 in Sauce or VirtualBox. Note no errors (there are existing warnings) and that the expandables function. Open in IE9 if you want and see there's an error around classList.
4. Link this branch into cfgov-refresh and open in IE9 and navigate to a key page and note expandables fall back to no-js version (e.g. on /blog filter).


cfarm:
Recommendation for this review: test the change in at least one of the affected pages [listed here](https://github.cfpb.gov/CFPB/hubcap/issues/109) in Sauce Labs to confirm the expandables gracefully degrade as expected.

Scotchester:
Shouldn't `package-lock.json` be updated in this PR?

anselmbradford:
Yes, let me regenerate. The problem with switching branches.

anselmbradford:
Also, I discovered this has a second, custom classList check at https://github.com/cfpb/capital-framework/blob/canary/src/cf-atomic-component/src/utilities/dom-class-list/index.js#L34-L44 I'll test and remove.

anselmbradford:
@Scotchester This is ready for review again.

anselmbradford:
> One weirdness is that when I run npm install in capital-framework, it's changing all of the resolved tarball URLs to https. I wonder why (1) they weren't already, and (2) your npm isn't making them so.

I've seen it convert them the other way before. Never understood why.

wpears:
The npm https -> http was a bug in npm itself 💇‍♂️ 
It's fixed in more recent versions of npm, but if anyone notices http packages getting added, they can be harmlessly find/replaced to https.

`sed -i '' -e 's/http:\/\//https:\/\//' package-lock.json` etc

PR #885, opened 2018-12-18
Add Yarn and Lerna to project, remove CI

contolini:
Adds Lerna for monorepo management and Yarn for it's neato workspaces and caching features.

Check out the [contributing docs](https://github.com/cfpb/capital-framework/blob/0e95345db34ad8e9a6349af8468ca4101b311292/CONTRIBUTING.md#changing-the-codebase) to see how these changes affect CF development. Spoiler alert: The only thing you'll have to do differently is remember to use `yarn` instead of `npm`.

### Notable Changes

1. The `src/` directory in the project's root has been renamed to `packages/`.
1. All the Travis CI stuff has been removed. We'll manually publish components (which is [easy with Lerna](https://github.com/cfpb/capital-framework/blob/7f1e16add97dd5abd91f6d9b72feba4ac149b057/CONTRIBUTING.md#release-management)) for the time being and we can move the task back into Travis if/when we're ready.
1. No more npm or `package-lock.json`. Node modules are cached in the git repo.
1. No more scary scripts for publishing/linking. Yarn and Lerna do most of it natively.

### Testing

1. Pull down this branch (might take a minute with all the cached packages).
1. `rm -rf node_modules/` to ensure you're starting with a clean slate.
1. `yarn` to install (no more `npm install`)
1. `yarn test` to confirm the tests pass as normal.
1. `yarn run cf-link` to link all the components
1. Edit `/packages/cf-core/src/cf-vars.less` and set the text color on line 13 to something silly like `red`.
1. `cd ~/wherever/cfgov-refresh`
1. `yarn link cf-core`
1. `gulp clean && gulp build`
1. Start up cfgov-refresh and open http://0.0.0.0:8000/. Text should now be red, indicating that it's correctly picking up your local version of `cf-core`.

### Unanswered questions

1. It's currently configured to use [fixed](https://github.com/lerna/lerna#how-it-works) versioning. Should we use independent versioning instead? We've been split on this. Let's keep discussing. **Update:** We discussed this and are happy with fixed versioning.

PR #887, opened 2019-02-07
Stylelint update + update dependencies

anselmbradford:
Running `gulp lint:styles --fix` on canary results in an error currently. This fixes that error (an apostrophe in a comment) and updates the stylelint config and performs general linting fix across the codebase.

Also, updates babel and jest.

## Additions

- Adds `stylelint-config-standard` for base stylelint configuration.

## Removals

- Unused `jest-when` package.

## Changes

- Replaces `.stylelintrc` with `stylelint.config.js`, which allows code comments and is not a hidden file (which follows the direction config best practices like babel has gone).
- Autofixes stylelint errors (indentation, spacing, missing semicolons).

## Testing

1. `gulp lint:styles` should report no errors. `gulp lint:styles --fix` should leave files untouched.
1. Follow the [local testing instructions](https://github.com/cfpb/capital-framework/blob/canary/CONTRIBUTING.md#testing-components-locally) to run the documentation and see that all looks like production at https://cfpb.github.io/capital-framework/components/

## Note
Diff may be difficult to view because of the indentation changes. It may be easier to view the files in this PR vs on canary.


src/cf-layout/src/cf-layout.less
```diff
@@ -257,18 +255,22 @@
...
+        unit( @grid_gutter-width / @base-font-size-px, em )
+        unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
 
     .respond-to-min( @bp-sm-min, {
         .grid_column( 12 );
-        padding: unit( ( @grid_gutter-width * 1.5 ) / @base-font-size-px, em )
-                 unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
+
+        padding:
+        unit( ( @grid_gutter-width * 1.5 ) / @base-font-size-px, em )
```
anselmbradford: The apparent missing indent here and in the one below appears to be a shortcoming with stylelint, as it's trying to align it with padding options on L:259-260. It appears the respond-to-min mixin is throwing it off. Do we care? While this is slightly unfortunate, the other things the linter fixes in this PR outweighs the two rules minor indenting inconsistency here.

contolini:
Let's hold this and we can use it to test our new changelog system from #885!

src/cf-core/src/cf-utilities.less
```diff
@@ -229,41 +238,41 @@
 // Margin utilities
 //
 
-.u-mt0  { margin-top:    0 !important; }
-.u-mb0  { margin-bottom: 0 !important; }
-.u-mt5  { margin-top:    5px !important; }
-.u-mb5  { margin-bottom: 5px !important; }
-.u-mt10 { margin-top:    10px !important; }
+.u-mt0 { margin-top: 0 !important; }
```
Scotchester: Maybe have stylelint ignore the manual alignments in our margin and width utilities in this file?

src/cf-forms/src/atoms/form-alert.less
```diff
@@ -22,7 +22,6 @@
     }
 }
 
-
```
Scotchester: I see you're setting max-empty-lines to 2 in the config, but here and many other places, 2nd empty lines are being removed. Did you add the rule after you saw that, but not retry the auto-fixer?

src/cf-icons/src/cf-icons.less
```diff
@@ -396,12 +397,14 @@
 .cf-icon__rotate( @degrees, @rotation ) {
     // TOOD: Check whether `filter` can be replaced
     //       by `-ms-transform` for CF supported browsers.
-    filter: progid:DXImageTransform.Microsoft.BasicImage( rotation=@rotation );
+    /* stylelint-disable-next-line function-name-case */
+    filter: progid:dximagetransform.microsoft.basicimage( rotation=@rotation );
```
Scotchester: Did you mean to revert the auto-fix on the function name here and a few lines below?

Another approach (besides ignoring these lines) would be using the ignore option for this rule: https://stylelint.io/user-guide/rules/function-name-case/#ignorefunctions-regex-as-string-regex-non-regex

src/cf-layout/src/cf-layout.less
Scotchester: Depending on what our linting posture will be going forward, we could maybe revert the auto-fix in these cases. If we get everything known-good with this PR, could we then only lint changed lines going forward? Another option could be to not use the shorthand property here.

stylelint.config.js
```diff
@@ -0,0 +1,44 @@
...
+
+rule-empty-line-before -
+  Custom setting that differs from stylelint-config-standard.
+
+max-empty-lines -
+  Set to 2 since we have existing two line breaks in place. Could be removed.
+
+no-descending-specificity -
+  Turned off, but probably shouldn't be.
+  TODO: Turn on this rule and see if issues can be fixed.
```
Scotchester: I agree it'd be good to try this.

stylelint.config.js
```diff
@@ -0,0 +1,44 @@
...
+
+selector-list-comma-newline-after -
+  Turned off because it wraps arguments less mixin declarations.
```
Scotchester:
```suggestion
  Turned off because it wraps arguments in Less mixin declarations.
```

src/cf-core/src/cf-utilities.less
```diff
@@ -148,36 +148,45 @@
...
+@c:  @link-text,
+@v:  @link-text-visited,
+@h:  @link-text-hover,
+@f:  @link-text,
+@a:  @link-text-active,
+@bc: @link-underline,
+@bv: @link-underline-visited,
+@bh: @link-underline-hover,
+@bf: @link-underline,
+@ba: @link-underline-active
```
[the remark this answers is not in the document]
anselmbradford: That looks (logically) correct. Things inside `{ … }` would be indented to indicate nesting, however, this is the arguments inside `( … )`, which are at the same level as the mixin declaration.

src/cf-core/src/cf-utilities.less
anselmbradford: If the tradeoff is adding code comments to ignore stylelint rules versus having the visual alignment off by a space, I'd go with the visual misalignment. It's not like it's incomprehensible with the values not totally lining up, and having additional code comments seems like clutter that's even worse. Arguably particular lines of code should not depend on the formatting of the lines around them anyway (other than nesting).

src/cf-core/src/cf-media-queries.less
```diff
@@ -9,30 +9,36 @@
 
 .respond-to-min( @bp, @rules ) {
     @ems: unit( ( @bp / @base-font-size-px ), em );
-    @media only all and ( min-width: @ems ) {
+
+    @media only all and (min-width: @ems) {
```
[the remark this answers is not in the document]
anselmbradford: Fixed in https://github.com/cfpb/capital-framework/pull/887/commits/ed7c21f882a5e386bc538e80069dd29977abcf40

src/cf-forms/src/atoms/form-alert.less
anselmbradford: Hm, yeah I may have run it with the default config and then added max 2 empty lines and ran the auto-fixer, but that would only truncate 3 lines down to 2, not bring 1 line up to 2. Looking at this file though, 1 empty line seems fine for that spot. Do we have other spots where it would be preferable as 2 lines? The ones like cf-pagination are above code comment "headers" but I don't mind it as 1 line. The comments themselves are 3 lines tall.

src/cf-icons/src/cf-icons.less
anselmbradford: Oops, nope, fixed. Since this has a TODO to check replacing these filters I'll leave it as inline ignore code comments, since the whole block can likely be stripped out.

src/cf-layout/src/cf-layout.less
anselmbradford: Autofixing will look at all files, so if we submitted this known-good the next time autofixing was run it would move these lines again. I tried stylelint code comments to disable the linting for this section (`/* stylelint-disable */` `/* stylelint-enable */`), but it inexplicably ignores it. This seems to be a known issue, but not repeatable by the maintainers https://github.com/stylelint/stylelint/issues/456 
I looked at breaking out the shorthand here, but it does add a lot of extra lines here, so I'd lean toward living with the indenting here and revisiting in a future stylelint update and see if it can be ignored. Happy to go the breakout route though if that's preferable to the (mis)indent.

src/cf-icons/src/cf-icons.less
anselmbradford: Hmmmm okay discovered the stylelint ignore comments don't work here either :(
I've updated to use `function-name-case` as you suggested.

src/cf-core/src/cf-utilities.less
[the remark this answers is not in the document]
Scotchester: I think typically arguments would get indented as well, but regardless, it didn't give this treatment to the similar mixin arguments in the lines above this.

src/cf-forms/src/atoms/form-alert.less
Scotchester: I don't have a consistent philosophy for when it should happen, it's sort of a know-it-when-you-see-it situation.

PR #893, opened 2019-03-06
Fix lerna-changelog command

contolini:
Our git tags are in the format `1.2.3` not `v1.2.3`. Oopsie.

This PR is also being used to test our new changelog process.


anselmbradford:
Maybe we want to drop the `v`? Why do we have it?

contolini:
None of our git tags use `v` -- I had accidentally added it to the script.

PR #895, opened 2019-03-07
cf-notifications: Set a max-width

anselmbradford:
This incorporates the max-width added to a version of the notification via mixing in `m-full-width-text` https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/regulations3k/jinja2/regulations3k/notification.html#L22

## Changes

- Sets a max-width on the notification. 

## Testing

1. Follow the local testing instructions to view the changes in the docs. 

## Screenshots

Before:

cf-notifications-before

After:

notification-after

### DIFF
--- a/packages/cf-notifications/src/molecules/notification.less
+++ b/packages/cf-notifications/src/molecules/notification.less
@@ -9,6 +9,9 @@
     background: @notification-bg;
     border: 1px solid @notification-border;
 
+    // This magic number makes for line lengths between 85-95 characters.
+    max-width: 41.8em;
+
     .cf-icon-svg {
         position: absolute;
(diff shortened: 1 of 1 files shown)


anselmbradford:
> Edit: cf-layout uses 41.875rem. Should we move this magic number into a variable?

Hmmm we probably should in cf-core, and use 41.875, not 41.8.

anselmbradford:
After chatting with @Scotchester we've decided to close this in favor of having the parent container dictate the width of the notification. This PR was largely in service of the regs3k notification, but I addressed this another way in https://github.com/cfpb/cfgov-refresh/pull/4891

PR #900, opened 2019-03-28
cf-forms: Fix checkbox and radio button grid stuff in IE

Scotchester:
Autoprefixer was not correctly being skipped for the `display: inline-grid;` property on the labels of our radio buttons and checkboxes. The Less step was removing the [Autoprefixer control comment](https://github.com/postcss/autoprefixer#control-comments), so Autoprefixer didn't know to skip it. Adding the `!` to the comment opening characters tells Less to keep the comment in so Autoprefixer now sees it.

I also switched from the `/* autoprefixer: off */` method, which required that the rule to be skipped lived in its own block, to `/* autoprefixer: ignore next */`, so we can do away with those extra blocks.

## Changes

- **cf-forms:** Switch from `/* autoprefixer: off */` to `/*! autoprefixer: ignore next */` to correctly exclude the radio & checkbox label `display: inline-grid;` rule from Autoprefixer

## Testing

Follow the instructions on https://github.com/cfpb/capital-framework/blob/master/CONTRIBUTING.md and view a page with checkboxes in IE, for example, inside the filter controls expandable on https://www.consumerfinance.gov/about-us/blog/.

### Before

Screen Shot 2019-03-28 at 16 48 49

### After

Screen Shot 2019-03-28 at 16 49 05

## Checklist

### Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Internet Explorer 9, 10, and 11
- [x] Edge
- [x] iOS Safari
- [x] Chrome for Android

### DIFF
--- a/packages/cf-forms/src/molecules/form-fields.less
+++ b/packages/cf-forms/src/molecules/form-fields.less
@@ -11,6 +11,16 @@
     &__checkbox,
     &__radio {
         .a-label {
+            // We need to turn off autoprefixing for the inline-grid because
+            // IE does not handle an inline-grid like other browsers,
+            // leading to an extremely narrow column of text for the label
+            // and the checkbox or radio widget covering the first part of it.
+            // The Autoprefixer ontrol comment below ensures that the following
+            // property is only picked up by browsers with standard support.
+            // The excalamation mark is necessary for Less to preserve the
+            // comment so that Autoprefixer will see it.
+            /*! autoprefixer: ignore next */
+            display: inline-grid;
             // 30px is width of checkbox/radio button plus the needed padding.
             grid-template-columns: unit( 30px / @base-font-size-px, em ) auto;
             vertical-align: top;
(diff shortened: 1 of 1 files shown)


packages/cf-forms/src/molecules/form-fields.less
```diff
@@ -11,6 +11,16 @@
     &__checkbox,
     &__radio {
         .a-label {
+            // We need to turn off autoprefixing for the inline-grid because
+            // IE does not handle an inline-grid like other browsers,
+            // leading to an extremely narrow column of text for the label
+            // and the checkbox or radio widget covering the first part of it.
+            // The Autoprefixer ontrol comment below ensures that the following
```
anselmbradford: Noting for the next minor update: `ontrol` --> `control`

packages/cf-forms/src/molecules/form-fields.less
```diff
@@ -11,6 +11,16 @@
...
+            // property is only picked up by browsers with standard support.
+            // The excalamation mark is necessary for Less to preserve the
```
anselmbradford: `excalamation` --> `exclamation`

packages/cf-forms/src/molecules/form-fields.less
Scotchester: Nooooooooooo!

packages/cf-forms/src/molecules/form-fields.less
Scotchester: Ugh, I hate myself.

contolini:
I don't think it's a side effect of this PR but I'm posting it here in case it's related: In IE [helper text](https://cfpb.github.io/capital-framework/components/cf-forms/#inputs-helper-text) is inline with the label instead of below it.

21c8d980-5797-11e9-9c8f-29d5d6fefdd8

anselmbradford:
@contolini https://github.com/cfpb/capital-framework/pull/631#issuecomment-353662165

PR #915, opened 2019-04-23
cf-notifications: Adjust message/explanation and links margins

anselmbradford:
## Changes

- Adjust margins of links in notifications.

## Testing

1. Pull branch down.
3. Ensure you're in the docs folder with `cd docs`
2. Run `gulp build && gulp docs && bundle exec jekyll serve` (run `gem install bundler && bundle` if jekyll is not installed.)

## Screenshots

Screen Shot 2019-04-23 at 10 32 30 AM

---

Screen Shot 2019-04-23 at 10 32 42 AM

---
The below are what one link would look like, but this is just for show and docs have two links only.
---

Screen Shot 2019-04-23 at 10 32 04 AM

---

Screen Shot 2019-04-23 at 10 31 52 AM


anselmbradford:
> While you're in cf-notifications/usage.md, can you fix the TOC link? It's missing an s

Fixed in https://github.com/cfpb/capital-framework/pull/915/commits/ea42ec5155a1f72c0962d29668c1ae40b0e3d75d

PR #916, opened 2019-04-30
cf-notifications: Remove unused classes

anselmbradford:
`m-notification__default` is not a class that exists. `m-notification` on its own without a modifier IS the default.

## Removals

- Removes `m-notification__default` from the docs.


anselmbradford:
cfgov-refresh PR https://github.com/cfpb/cfgov-refresh/pull/4977

PR #919, opened 2019-05-07
Add missing icons

anselmbradford:
The doc templates include some icons, but the CSS ALSO references some icons. The includes are in `_includes`, but the CSS references `dist`, so we need an additional copy there. It would be nicer if this didn't fail silently when the icons are missing and it would also be nice not to require duplicate copies.

## Additions

- Approved and down icons that cf-forms checkboxes and selects use, respectively.

## Testing

1. Pull this branch down.
1. /components/cf-forms/ should have checks in the checkboxes, and a down arrow in the select menus.


anselmbradford:
> Hmm... is this only a problem with the docs? We probably need to think harder about how these icons get where they need to go if someone is adding cf-forms to a new project.

Turns out this is a problem when ccdb5-ui tries to update cf-forms too! 👀

PR #930, opened 2019-06-06
Update Featured Content Module styling

Scotchester:
This PR implements several desired styling updates for Featured Content Modules. See https://GHE/CFGOV/platform/issues/3087 for background.

/cc @designlanguage @keelerr

## Removals

- Remove category headers from FCM templates

## Changes

- Change FCM min height from 320px to 220px
- Set FCM image width to 270px, let text flex
- Change FCM breakpoint to 600px
- Add bottom padding to FCM image on XS screens
- Sets border on the FCM itself, instead of surrounding block
  - The border is inherent to the FCM styling and shouldn't be set on the block. It makes even less sense now that the block modifier for adding the negative side margins is no longer used.
- Updates docs

## Testing

1. Check out a working branch
1. Run `gulp docs`
1. Go into the documentation directory with `cd docs`
1. Run `bundle exec jekyll serve` and visit http://localhost:4000/components/cf-layout/#featured-content-module
1. Go into responsive mode and view the FCM at various window widths

### Before

Screen Shot 2019-06-06 at 13 12 40

### After

Screen Shot 2019-06-06 at 13 13 33

## Todos

- [ ] Open corresponding PR in cfgov-refresh

## Checklist

### Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] iOS Safari
- [x] Chrome for Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS


docs/_includes/usage/cf-layout/usage.md
```diff
@@ -1290,208 +1290,181 @@ height of the next parent using `position: relative;`. This means that the
...
-{% endcomment %}
 
-Featured content module, like a hero, consists of headline/text/optional call
-to action along with a visual. It is intended to be used in a main content
-column next to a sidebar.
+Featured content modules, like a hero, consist of headline and description,
+an optional call to action, and a static or video visual.
+They are intended to be used in a main content column next to a sidebar.
 
 Text is full width & displayed above the visual in the default/mobile view.
```
anselmbradford: Minor nitpick, but why are we using ampersands here? Probably would be stylistically better to spell out `and`.

docs/_includes/usage/cf-layout/usage.md
```diff
@@ -1290,208 +1290,181 @@ height of the next parent using `position: relative;`. This means that the
...
-        <div class="o-featured-content-module_visual">
-            <img class="o-featured-content-module_img"
-                 src="http://placekitten.com/g/540/270">
-        </div>
-    </section>
-</div>
+At larger screen sizes, the image moves to the right of the text,
+occupying a fixed width of 270px (equal to 3 of 12 columns at max page width).
+When creating a static image for the visual, it should be 1076px × 606px
+(a 16:9 aspect ratio), which is 2x the rendered with at the maximum size
```
anselmbradford: `which is 2x the rendered with` … should be `width` not `with`.

docs/_includes/usage/cf-layout/usage.md
```diff
@@ -1290,208 +1290,181 @@ height of the next parent using `position: relative;`. This means that the
...
+at which the visual will be seen (on a 600px wide display).
+
+By default, the image is anchored on its left when the FCM is viewed on a
```
anselmbradford: Should it say `anchored on its center-left` or `anchored on its middle-left`, as that's what it looks like it is doing?

docs/_includes/usage/cf-layout/usage.md
```diff
@@ -1290,208 +1290,181 @@ height of the next parent using `position: relative;`. This means that the
...
-        </div>
-        <div class="o-featured-content-module_visual">
-            <img class="o-featured-content-module_img"
-                 src="http://placekitten.com/g/540/270">
-        </div>
-    </section>
-</div>
+### Anchoring the visual on the right
+
+When one wants to anchor a featured content module's visual on the right
```
anselmbradford: Could be shortened to `To anchor a featured content module's visual on the right…`

docs/_includes/usage/cf-layout/usage.md
```diff
@@ -1290,208 +1290,181 @@ height of the next parent using `position: relative;`. This means that the
...
+(for example, when displaying a map with a copyright watermark on the right),
+add the `o-featured-content-module__right` modifier class
+to the `o-featured-content-module` organism.
+This anchors the right side of the image to the right of the available space
```
anselmbradford: Should it be `right-middle` or `right-center`?

docs/_includes/usage/cf-layout/usage.md
```diff
@@ -1290,208 +1290,181 @@ height of the next parent using `position: relative;`. This means that the
...
-        </div>
-        <div class="o-featured-content-module_visual">
-            <img class="o-featured-content-module_img"
-                 src="http://placekitten.com/g/540/270">
-        </div>
-    </section>
-</div>
+At larger screen sizes, the image moves to the right of the text,
+occupying a fixed width of 270px (equal to 3 of 12 columns at max page width).
+When creating a static image for the visual, it should be 1076px × 606px
```
anselmbradford: Says image should be 1076×606, but the example images used are 540×270. Should this be noted?

docs/_includes/usage/cf-layout/usage.md
Scotchester: I tried to clarify this below in the right-anchored description (and I should have written it the same way here), but it looks like that was confusing to you also, so perhaps I should go back to the drawing board 😆 

How does this sound (including some minor edits to the previous paragraph)?

---

Text is full width and displayed above the visual in the default/mobile view.
When creating a static image for the visual, it should be 1076px × 606px
(a 16:9 aspect ratio), which is 2x the rendered width at the maximum size
at which the visual will be seen (on a 600px wide display).

At larger screen sizes, the image moves to the right of the text,
occupying a fixed width of 270px (equal to 3 of 12 columns at max page width).
By default, the left edge of the image is anchored to the left side of the
visual's 270px-wide area, and the right side is cropped off.
The height of the visual area decreases as screen size increases,
resulting in slightly different image cropping at different screen sizes,
but the left edge of the image remains anchored in view.
See below for modifiers that change the image anchoring.

docs/_includes/usage/cf-layout/usage.md
Scotchester: :+1: Updated in 43458fae.

docs/_includes/usage/cf-layout/usage.md
Scotchester: Good catch; thanks! Fixed in 43458fae.

docs/_includes/usage/cf-layout/usage.md
Scotchester: Updated in 43458fae.

docs/_includes/usage/cf-layout/usage.md
Scotchester: :+1: Updated in 43458fae.

docs/_includes/usage/cf-layout/usage.md
Scotchester: Updated all the placeholder images to be 1076x606. This also conveniently gives us an image where the kitten is more centered, which makes the different anchoring more obvious.

docs/_includes/usage/cf-layout/usage.md
anselmbradford: > The height of the visual area decreases as screen size increases,
resulting in slightly different image cropping at different screen sizes,
but the left edge of the image remains anchored in view.

Still not totally clear how the cropping works without looking at it closer. The top AND bottom get cropped as screen size increases? Or is it anchored to upper-left and the bottom only gets cropped?

anselmbradford: Also, should it be `Text is full-width …`? Not sure what we use elsewhere.

Scotchester: > Still not totally clear how the cropping works without looking at it closer. The top AND bottom get cropped as screen size increases? Or is it anchored to upper-left and the bottom only gets cropped?

The full height of the image remains visible; top and bottom are not cropped at all. I can add that in.

This is extraordinarily difficult to describe in words. The best thing to do is open the responsive design mode and play around with it. Maybe I should literally say that in the docs, too :)

> Also, should it be `Text is full-width …`? Not sure what we use elsewhere.

This phrase is leftover from the previous iteration of the docs. It is probably assumed to be true when one is talking about the mobile view, so it may not be necessary to state it that way. How about... "The text and visual are stacked vertically at screen widths of 600px and under."?

packages/cf-layout/src/organisms/featured-content-module.less
```diff
@@ -19,25 +19,31 @@
         padding-top: unit( @grid_gutter-width / @base-font-size-px, em );
         padding-bottom: unit( @grid_gutter-width / @base-font-size-px, em );
 
+        .respond-to-max( @bp-xs-max, {
+            padding-right: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
```
anselmbradford: TRIVIAL: This and the line below have extra parens you don't need. `unit( @grid_gutter-width / 2 / @base-font-size-px, em );` should work.

packages/cf-layout/src/organisms/featured-content-module.less
```diff
@@ -19,25 +19,31 @@
...
-            width: 300px; // Width of 5 of 12 cols at lg size
-            float: left;
         } );
     }
 
     &_visual {
+        .respond-to-max( @bp-xs-max, {
+            padding-right: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
+            padding-bottom: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
+            padding-left: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
```
anselmbradford: Three lines with extra parens. I'll leave it up to you if you want to remove them. We have both formats in the wild I think.

Scotchester:
@anselmbradford Thanks for the review! I believe I have addressed all your feedback. Ready for re-review.

P.S.: I still want to do a little more cross-browser testing. I see an issue in IE9 already 😩

packages/cf-layout/src/cf-layout.less
```diff
@@ -45,8 +45,8 @@
 
 @fcm-bg:                        @block__bg;
 
-@fcm-bp-max:                    700px;
-@fcm-bp-min:                    701px;
+@fcm-bp-max:                    600px;
+@fcm-bp-min:                    601px;
```
anselmbradford: Do we need these any more or could we use the standard breakpoints directly? https://github.com/cfpb/capital-framework/blob/ad6871887c3d851b6dfe34fd6dd36cf63e27c843/packages/cf-core/src/cf-vars-breakpoints.less#L9-L10

packages/cf-layout/src/organisms/featured-content-module.less
```diff
@@ -19,25 +19,31 @@
...
 
+        .respond-to-max( @bp-xs-max, {
+            padding-right: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+            padding-left: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+        } );
+
         .respond-to-min( @fcm-bp-min, {
-            padding-right: unit( @grid_gutter-width / @base-font-size-px, em );
+            flex: 1;
+            padding-right: 300px;
```
anselmbradford: Just checking, fixed padding here is correct?

packages/cf-layout/src/cf-layout.less
Scotchester: Talked to Chris to ensure he didn't intend something that I was unaware of. He did not, so I cleaned up the variables.

packages/cf-layout/src/organisms/featured-content-module.less
Scotchester: It is, but it should be calculated from variables. Updated in the forthcoming commit.

Scotchester:
I think this is good to go. Fixed the IE9 bug by adding an explicit `top: 0;` to the visual's absolute positioning. @anselmbradford, any final comments?

PR #931, opened 2019-06-12
cf-layout: Fix issues in docs

anselmbradford:
## Removals
- Remove reference to non-existant `content_bar` class.
- Remove redundant block link in cf-layout table of contents.

## Changes

- Update outdated brand-colors link.
- Correctly nest cf-layout table of contents links.
- Move "Bleedbar sidebar styling" to correct section of cf-layout table of contents.

## Testing

1. `cd docs && gulp docs && bundle exec jekyll serve`
2. Brand colors link should work in http://localhost:4000/capital-framework/components/cf-layout/#color-variables
3. TOC links should work on http://localhost:4000/capital-framework/components/cf-layout/

## Todos

- The cf-grid sections in cf-layout docs seem like they should move to cf-grid.

### DIFF
--- a/packages/cf-layout/usage.md
+++ b/packages/cf-layout/usage.md
@@ -130,9 +130,6 @@ When using the modifiers described below to create columns,
 the columns will remain stacked for smaller screens and then convert to to
 columns at `801px`.
 
-`.content_line` must come after `.content_hero` (if it exists) but before
-`.content_wrapper`.
-
 _Inline styling is for demonstration purposes only; do not include it in your
 markup._
 
@@ -140,7 +137,6 @@ markup._
     <section class="content_hero" style="background: #E3E4E5">
         Content hero
     </section>
-    <div class="content_line"></div>
     <div class="content_wrapper">
         <section class="content_main">
             Main content area
@@ -156,7 +152,6 @@ markup._
     <section class="content_hero" style="background: #E3E4E5">
         Content hero
     </section>
-    <div class="content_line"></div>
     <div class="content_wrapper">
         <section class="content_main">
             Main content area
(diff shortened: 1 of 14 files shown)


Scotchester:
I'm not sure how much we should invest in this with the Design System coming down the pike. Documenting cf-layout and cf-grid coherently is quite difficult. That said, if we want to try and push this through...

> Correctly nest cf-layout table of contents links.

It was actually mostly "correct" before. The heading "Content layouts" refers specifically to the `content-l` class (block, in BEM parlance) and its related elements and modifiers. The "Content layout column dividers" section should be moved under "Content layouts", but everything following is not a part of that set of patterns.

> Move "Bleedbar sidebar styling" to correct section of cf-layout table of contents.

As this only applies to 2:1 layouts, I'd say this section should be made a child of the "Right-hand sidebar layout" section.

> The cf-grid sections in cf-layout docs seem like they should move to cf-grid.

Well, the docs are documenting where that code lives. One could argue that code should move from cf-layout to cf-grid, though.

packages/cf-layout/usage.md
```diff
@@ -432,7 +130,7 @@ When using the modifiers described below to create columns,
 the columns will remain stacked for smaller screens and then convert to to
 columns at `801px`.
 
-`.content_bar` must come after `.content_hero` (if it exists) but before
+`.content_line` must come after `.content_hero` (if it exists) but before
```
Scotchester: I think this sentence, and the two `<div class="content_line"></div>` just below, can be deleted.

anselmbradford: Removed in https://github.com/cfpb/capital-framework/pull/931/commits/ec5377ff8fc3a9b0a0b189e0b2575ad27cb7caba

PR #934, opened 2019-06-17
cf-forms: Fix issue where path to icons was hardcoded

anselmbradford:
## Changes

- Fix issue where path to icons was hardcoded for form-fields and select drop-downs.

## Testing

2.  Visit http://localhost:4000/components/cf-forms and see the checkboxes have checks and selects have a drop-down arrow. 
3. You can change the value for `@cf-icon-path` and re-run the command above to see that the icons break.

### DIFF
--- a/packages/cf-forms/src/atoms/select.less
+++ b/packages/cf-forms/src/atoms/select.less
@@ -64,7 +64,7 @@
         background-color: @select-icon-bg;
         content: '';
         pointer-events: none;
-        background-image: url( '../icons/down.svg' );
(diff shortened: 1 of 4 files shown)


packages/cf-forms/src/atoms/select.less
```diff
@@ -64,7 +64,7 @@
         background-color: @select-icon-bg;
         content: '';
         pointer-events: none;
-        background-image: url( '../icons/down.svg' );
```
jimmynotjim: Lol, this was probably my doing. Sorry

PR #937, opened 2019-06-21
Minor formatting and code comment fixes

anselmbradford:
## Changes

- Fix typo in code comment.
- Add explanation of what CSS content rule is adding.
- Add spaces inside parentheses for consistency with the rest of the code.

## Testing

1. `gulp` should pass.


packages/cf-grid/src/cf-grid.less
```diff
@@ -12,10 +12,10 @@
 // Less variables
 //
 
-@grid_wrapper-width:            1230px;
-@grid_gutter-width:             30px;
-@grid_total-columns:            12;
-@grid_debug:                    false;
+@grid_wrapper-width: 1230px;
```
anselmbradford: Matches formatting in https://github.com/cfpb/capital-framework/blob/master/packages/cf-grid/src-generated/cf-grid-generated.less#L27

packages/cf-typography/src/cf-typography.less
```diff
@@ -24,35 +24,35 @@
 // Running text elements
 
 // .a-micro-copy
-@micro-copy:                    @black;
+@micro-copy: @black;
```
anselmbradford: Matches indenting of variables at the top of the file https://github.com/cfpb/capital-framework/blob/master/packages/cf-typography/src/cf-typography.less#L11

anselmbradford:
Thanks @wpears !

PR #941, opened 2019-06-24
Add design system deprecation notice

contolini:
Adds a deprecation notice to the readme, pull request template and component documentation pages.

LMK if you think the message should be changed.

Screen Shot 2019-06-24 at 11 31 19 AM

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,5 @@
+**Heads up:** In an effort to unify our design resources, we will disable this repo and migrate Capital Framework's code to [cfpb/design-system](https://github.com/cfpb/design-system) in Summer 2019. Until the migration occurs, please continue to use this repo.
+
 [Short description explaining the high-level reason for the pull request]
 
 ## Additions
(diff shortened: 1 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,5 @@
+**Heads up:** In an effort to unify our design resources, we will disable this repo and migrate Capital Framework's code to [cfpb/design-system](https://github.com/cfpb/design-system) in Summer 2019. Until the migration occurs, please continue to use this repo.
```
anselmbradford: I'd omit this from the pull request template, since it'll be something we have to remove relatively soon, and in general it would be a good idea to keep the template consistent across projects. If for some reason someone opens a pull request in conflict with the migration, a comment could be left to point them to this notice in the readme.

anselmbradford: Oh wait, I'd misread the "disable this repo" part. I guess it doesn't really matter in that case.

contolini: We'll archive it and I think that prevents PRs

PR #942, opened 2019-06-27
Improve jump link docs

Scotchester:
I've had these minor doc improvements kicking around a local branch for ages.

## Changes

- Clarifies block link and jump link docs

## Checklist


docs/_includes/usage/cf-typography/usage.md
```diff
@@ -299,8 +298,28 @@ For more information, email
 </a>.
 ```
 
+### Block link
```
anselmbradford: Should this be `Block link modifier`?

Scotchester: I suppose it technically could be, but it's never used in markup as a modifier class.

anselmbradford: What do you mean? The class is `a-link__block`

docs/_includes/usage/cf-typography/usage.md
```diff
@@ -299,8 +298,28 @@ For more information, email
...
+
+The block link class is used as a mixin in the [jump link](#jump-link) and
+[link list](#link-list-modifier) patterns to convert a standard inline link to
+a block-level element with `padding`, `background`, and `border` at extra-small
+screen sizes. It is not intended to be used directly.
+
+<a class="a-link__block" href="#">
```
anselmbradford: Should be `a-link a-link__block` if we're following the pattern elsewhere, like jump links below.

docs/_includes/usage/cf-typography/usage.md
Scotchester: As noted in the wording I'm adding in this PR, it's not currently ever used directly by writing it into the markup, and probably shouldn't be. It's mixed-in to jump link and link lists links.

docs/_includes/usage/cf-typography/usage.md
Scotchester: I wonder if we should remove this example and just let people look at the jump link example.

anselmbradford: Yeah, seems like it should follow like "Jump link with large link modifier" as "Jump link with block link modifier"

docs/_includes/usage/cf-typography/usage.md
anselmbradford: Ah gotcha, yeah that should be clearer in the actual example.

Scotchester:
@anselmbradford Pushed another commit to try to further clarify the block link modifier mixin. LMK what you think.

anselmbradford:
> @anselmbradford Pushed another commit to try to further clarify the block link modifier mixin. LMK what you think.

Don't forget to run `gulp docs` and push the changes.

packages/cf-typography/usage.md
```diff
@@ -23,8 +23,8 @@ and has more basic typography patterns.
     - [Meta header](#meta-header)
 - [Link patterns](#link-patterns)
     - [Links with icons](#links-with-icons)
-    - [Jump link](#jump-link)
     - [Block link](#block-link)
+    - [Jump link](#jump-link)
```
anselmbradford: Order is wrong here. Jump links appear first.

anselmbradford:
Could we eliminate `a-link__block` entirely? I can't find a location it's used. Do you know of one?

packages/cf-typography/usage.md
Scotchester: Fixed in amended commit by eliminating the block link entry from the TOC entirely. Since it's not something that should be used directly, one shouldn't need to be looking for it in the TOC.

Scotchester:
> Don't forget to run `gulp docs` and push the changes.

D'oh! Fixed, amended, force-pushed.





> Could we eliminate `a-link__block` entirely? I can't find a location it's used. Do you know of one?

Again, it is not used in markup class lists, only as a mixin to [jump links](https://github.com/cfpb/capital-framework/blob/ded88e79ccedfdca38e44698e8a7a005509f7094/packages/cf-typography/src/atoms/links.less#L32) and [link lists](https://github.com/cfpb/capital-framework/blob/ded88e79ccedfdca38e44698e8a7a005509f7094/packages/cf-typography/src/molecules/list.less#L63).

Perhaps what we should do is rename and move it so that it does not appear to be a modifier of `a-link`? `.u-block-link` or something?

anselmbradford:
> Again, it is not used in markup class lists, only as a mixin to jump links and link lists.

Oh! Forgive me for the sloppy searching. I was forgetting that a mixin did not mean it was a class added as a sibling to `a-link` and was searching around for that. So you're probably right a `u-` utility makes more sense.

**However,** do we need to document this at all? This appears to be a convenience mixin for avoiding repeating a large block of rules in two locations and is only used internally. And, if I'm not mistaken is intended to only be used internally? The CF docs should probably be left to the public API only and convenience code blocks like this could have a comment within the code itself saying what it's used for.

Scotchester:
> Oh! Forgive me for the sloppy searching. I was forgetting that a mixin did not mean it was a class added as a sibling to `a-link` and was searching around for that. So you're probably right a `u-` utility makes more sense.
> 
> **However,** do we need to document this at all? This appears to be a convenience mixin for avoiding repeating a large block of rules in two locations and is only used internally. And, if I'm not mistaken is intended to only be used internally? The CF docs should probably be left to the public API only and convenience code blocks like this could have a comment within the code itself saying what it's used for.

Yeah, I think that would be fine. I think this situation is a result of previously having documented everything in code comments and using Topdoc to produce a docs site out of that, and when we moved away from Topdoc, we went all the way in the opposite direction – leaving line-level comments in code, but taking all higher-level conceptual docs out of the code. I think we've proven here that, as is so often the case, the solution lies somewhere in the middle. I'll be back in a bit with that update.

Scotchester:
Updated, rebased on master, force-pushed! Ready for final review, @anselmbradford.

PR #950, opened 2019-07-23
cf-forms: Remove unused form input variables | add missing one

anselmbradford:
## Removals

- Remove unused `@input-border__active` and `@input-border__selected` variables.

## Changes

- Converts `#fff` to a variable like the other colors.

## Testing

1. `cd docs && gulp build && gulp docs && bundle exec jekyll serve`

## Notes

- Possibly we should use the color variables directly, so `@white` instead of `@form-field-input-border-inset__selected`.

### DIFF
--- a/packages/cf-forms/src/molecules/form-fields.less
+++ b/packages/cf-forms/src/molecules/form-fields.less
@@ -157,19 +157,21 @@
 
             &:checked + .a-label:before {
                 background-color: @input-bg__selected;
-                box-shadow: inset 0 0 0 2px @form-field-input-border-inset__selected;
+                box-shadow: inset 0 0 0 2px @form-field-inset;
             }
 
             &:focus:checked + .a-label:before,
             &.focus:checked + .a-label:before {
                 border-color: @input-border__focused;
-                box-shadow: 0 0 0 1px @input-border__focused, inset 0 0 0 2px @form-field-input-border-inset__selected;
+                box-shadow: 0 0 0 1px @input-border__focused,
+                            inset 0 0 0 2px @form-field-inset;
             }
 
             &:hover:checked + .a-label:before,
             &.hover:checked + .a-label:before {
                 border-color: @input-border__hover;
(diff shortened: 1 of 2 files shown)


packages/cf-forms/src/molecules/form-fields.less
```diff
@@ -157,19 +157,19 @@
...
                 border-color: @input-border__focused;
-                box-shadow: 0 0 0 1px @input-border__focused, inset 0 0 0 2px #fff;
+                box-shadow: 0 0 0 1px @input-border__focused, inset 0 0 0 2px @form-field-input-border-inset__selected;
             }
 
             &:hover:checked + .a-label:before,
             &.hover:checked + .a-label:before {
                 border-color: @input-border__hover;
-                box-shadow: 0 0 0 1px @input-border__hover, inset 0 0 0 2px #fff;
+                box-shadow: 0 0 0 1px @input-border__hover, inset 0 0 0 2px @form-field-input-border-inset__selected;
```
Scotchester: Hmm... maybe this should just be named `@input-border-inset`? Dropping the `form-field` would make it more consistent with `@input-border__focused` and `@input-border__hover`, and since it is always this same color regardless of state, the `__selected` modifier seems superfluous.

anselmbradford: I agree the names are long, however it appears `@input…` may be used on all kinds of inputs, while `@form-field` is only radio/checkboxes. Not sure if we want to stick with that. Maybe `@form-field-border-inset`? It seems like we _could_ drop `-input-` from all of the `@form-field-` variables.

Scotchester: Hmm, fair point about `@input`.

Another thing that occurs to me is that it isn't really part of the border, either technically or conceptually in how one would describe it when looking at it. So maybe just `@form-field-inset`?

anselmbradford: Works for me. Added in https://github.com/cfpb/capital-framework/pull/950/commits/c28e62c8ae3fa365d716e0aebef4012b4c0c1552

PR #951, opened 2019-07-23
cf-tables: Add basic table format with caption

### LINKED ISSUE #888: cf-tables: missing styling for <caption>
cf-tables does not style [`<caption>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption). By default there will be no padding between a table caption and it's table in cf-tables. Example of table caption use appears at http://consumerfinance.gov/credit-cards/agreements/issuer/alaska-usa-federal-credit-union/

anselmbradford:
Fixes #888

## Additions
- Add "basic" `o-table` section to docs.
- Add initial table `<caption>` styles.
- Add `@table-head-text` variable.

## Removals
- Remove `@thead-text` variable
- Remove `@thead-bg` variable and use `@table-head-bg`.

## Testing
1. `cd docs && gulp docs && gulp build && bundle exec jekyll serve`
2. Visit http://localhost:4000/components/cf-tables/#basic-tables

## Screenshots

Screen Shot 2019-07-23 at 12 55 58 PM


packages/cf-tables/src/cf-tables.less
```diff
@@ -35,6 +35,11 @@
 }
 
 .o-table {
+    caption {
```
Scotchester: This should move to cf-core with other base table styles.

anselmbradford: Force pushed and updated in https://github.com/cfpb/capital-framework/pull/951/commits/0c4efd125be7c54f56d19eb868fb3b44a93b39c7

anselmbradford: Before:
Screen Shot 2019-07-23 at 3 42 06 PM

After:
Screen Shot 2019-07-23 at 3 42 01 PM

PR #952, opened 2019-07-23
Style headings within table header cells only as standard table headers

Scotchester:
We have a page in which we'd like to use `h5` elements inside of a `th`. See https://GHE/CFGOV/platform/issues/2703 for background.

The issue is that, currently, putting a heading in a `th` messes with the font size and weight and brings in unwanted margin. This PR addresses that.

## Additions

- Adds rules to override standard heading styling for headings within `thead th` elements

## Testing

1. Pull branch
1. `gulp`
1. `cd docs && bundle exec jekyll serve`
1. Visit http://localhost:4000/components/cf-tables/
1. Scroll down to a table, inspect a header cell, edit its HTML, and change it so that its text is enclosed in a heading element
1. See that the styling does not change

## Checklist

### Browsers

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] Internet Explorer 9, 10, and 11
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS


packages/cf-core/src/cf-base.less
```diff
@@ -483,6 +483,19 @@ tbody tr {
 th {
     font-weight: 600;
     text-align: left;
+
```
anselmbradford: Let's add a code comment here. Superficially it seems like the real issue is that heading elements shouldn't be placed in table headings, however, they are needed for accessibility semantics as I understand? So maybe say:

```
// Heading elements may at times appear inside th elements as required for navigating the page's content with screenreaders. 
```

It may also be worth adding a note in the cf-tables usage doc for this as well.

Scotchester: Good call.

Scotchester: @anselmbradford Updated. LMK what you think.

docs/_includes/usage/cf-core/usage.md
```diff
@@ -1134,6 +1134,10 @@ Links within a `nav` element are not underlined.
 </table>
 ```
 
+_Note: While `th` elements normally only contain raw text, they may sometimes
```
anselmbradford: In a perfect world we'd be consistent with our notes in CF, but this is fine till design manual party time.

PR #955, opened 2019-07-24
cf-forms: Add disabled/checked states to form inputs

anselmbradford:
## Additions

- Adds styles for disabled/checked input states.
- Adds `@input-bg__disabled-selected` variable.
- Adds `a-select__disabled` class.

## Changes

- Allows passing of SVG background image fill color to `u-svg-inline-bg` plugin.

## Testing

2. Visit http://localhost:4000/components/cf-forms/ and compare to https://cfpb.github.io/capital-framework/components/cf-forms/

## Screenshots

Screen Shot 2019-07-24 at 11 56 40 AM
Screen Shot 2019-07-24 at 11 56 47 AM
Screen Shot 2019-07-24 at 11 56 53 AM
Screen Shot 2019-07-24 at 11 57 00 AM
Screen Shot 2019-07-24 at 12 41 07 PM


packages/cf-icons/src/cf-icons-svg-inline.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
 
     /**
      * @param {string} svgName - The canonical name of the icon.
-     * @param {string} svgFillColor - The fill color of the icon in rgb() format (defaults to black).
+     * @param {string} svgFillColor - The fill color of the icon (defaults to CFPB Black).
```
anselmbradford: We want to default to CFPB black for the icons? I'm not sure where CFPB versus pure black is used—but if it's an icon with black text it should be CFPB black?

Scotchester: The only non-legacy places where we use pure black are in the video overlay (where we thought a not-pure-black background alongside a video could be distracting) and in print styles, where pure black is supposed to print faster.

anselmbradford: Ah interesting ok

anselmbradford:
@Scotchester The disabled checkbox mark looks too dark to me, but if that fits with our design and contrast requirements this PR is 👍 for me. Feel free to merge or approve and I'll merge.

PR #956, opened 2019-07-26
cf-icons: Change reference to filename

anselmbradford:
## Changes

- Change how filename is referenced in less inline SVG plugin.


anselmbradford:
Merging as this is breaking cfgov-refresh

PR #959, opened 2019-08-15
Update dependencies, including linters, and fix linter errors

anselmbradford:
Babel updates are brought up to the version that are cfgov-refresh.

## Changes

- Updates `@babel/core`, `@babel/preset-env`, `gulp-eslint`, `gulp-ignore`, `lerna`, `stylelint`, `svgo`, and `terser-webpack-plugin`.
- Fixes stylelint errors.
- Add ignore to stylelint config for `@plugin` syntax in less.
- Updates eslint config for ESLint 6.
- Removes `font-family-no-missing-generic-family-keyword` stylelint setting.

## Testing

1. Pull branch, `gulp lint` should pass without errors (there will be warnings).
2. `cd docs && gulp && bundle exec jekyll serve` and http://localhost:4000/components/cf-forms/ should look the same as https://cfpb.github.io/capital-framework/components/cf-forms/


packages/cf-forms/src/molecules/form-fields.less
```diff
@@ -174,15 +175,17 @@
             &:focus:checked + .a-label:before,
             &.focus:checked + .a-label:before {
                 border-color: @input-border__focused;
-                box-shadow: 0 0 0 1px @input-border__focused,
-                            inset 0 0 0 2px @form-field-inset;
+                box-shadow:
+                    0 0 0 1px @input-border__focused,
+                    inset 0 0 0 2px @form-field-inset;
```
Scotchester: Instead of changing our customary format for properties like this, what do you think about adding `'declaration-colon-newline-after': null` to the rules list and modifying the `indentation` rule to:

```js
    'indentation': [
      4,
      {
        ignore: 'value'
      }
    ],
```

packages/cf-icons/src/cf-icons.less
```diff
@@ -23,12 +23,14 @@
 // Mixin to inject an SVG from the ./icons/ directory
 // into a background-image property.
 @plugin "cf-icons-svg-inline";
+
```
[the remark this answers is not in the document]
anselmbradford: That's my understanding.

packages/cf-tables/src/cf-tables.less
```diff
@@ -83,6 +83,7 @@
             content: '';
             visibility: hidden;
             .u-svg-inline-bg( 'down' );
+
```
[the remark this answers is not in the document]
anselmbradford: Done!

packages/cf-forms/src/molecules/form-fields.less
anselmbradford: 👍 Sounds fine to me. Done!

anselmbradford: I've removed `font-family-no-missing-generic-family-keyword` also, since it's not applicable now that the webfont icons are gone.

